PR #9, opened 2019-07-18
Ankylosing Spondylitis Disease Activity Score (ASDAS)

martinburchell:
I think this is done. I notice that https://www.asas-group.org/clinical-instruments/asdas-calculator/ does not allow a CRP value < 2mg/L and fixes to 2mg/L in this case. I have not done this.

### Checklist for new tasks:

- [X] Run self_test on server


martinburchell:
Secret strings commit ref ddfef7d7ef2b864a206a64b65fab948a7789c382

martinburchell:
Comments from @RudolfCardinal via email:

     o I wondered initially if it would be clearer for researchers to
       name the inflammatory fields "crp" and "esr", or "crp_mg_l" and
       "esr_mm_h" (also given that it becomes non-sequential for the
       ESR version, i.e. questions 1-4 and 6). However, your way is
       also very clearly set out; let's go with your system.
     o There is an additional anchor point for Q2 (5 = "1 [hour]"). I
       can see this is a pain to do. One option is to ignore it;
       another is to incorporate it into the labels -- e.g. labels "0
       (0 hours)", "1", "2", ... "5 (1 hour)", "6", ... "10 (2 or more
       hours)". Not sure how that would look. Another option would be
       to adapt the slider widget for two separate sets of labels;
       obviously more faff still. What's your feeling?
     o I've added the end anchor points as hints in the questions.
     o I've made a small tweak (adding an Optional[...] to the CRP/ESR
       scoring functions) to pass the PyCharm type checker.
     o good choice of tracker limits -- I have notched it up very
       slightly to 7.5 (I managed to get 7.08 with a CRP of 400 which
       is not impossible)

PR #11, opened 2019-07-22
ASDAS: Add upper limits for CRP and ESR

martinburchell:
CRP <= 2000, ESR <= 300 should cover all anticipated values
The means we no longer need the custom hint

@RudolfCardinal We're only enforcing this on the client side. Does that seem reasonable? Could easily add validators on the server side. Do we need to care about a minimum value for CRP? https://www.asas-group.org/clinical-instruments/asdas-calculator/ has a minimum of 2.0mg/L. We could do the same or leave it to the discretion of the clinician.

### DIFF
--- a/docs/source/tasks/asdas.rst
+++ b/docs/source/tasks/asdas.rst
@@ -54,7 +54,9 @@ Each question is weighted:
 Intellectual property rights
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-**TO BE ESTABLISHED**
+**TO BE ESTABLISHED.** Original paper (Lukas et al. 2009) is copyright 2009 BMJ
+Publishing Group and European League Against Rheumatism, but status of task
+itself less clear.
 
 
 History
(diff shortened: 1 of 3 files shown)


RudolfCardinal:
Seems reasonable to me to enforce this only on the client side. CRP values below 2 mg/L are perfectly plausible; e.g. https://www.ncbi.nlm.nih.gov/pubmed/25191699 . The "log(CRP+1)" prevents it blowing up mathematically as long as CRP is non-negative. So I don't think we need to worry about a lower limit. Similarly for ESR (not sure what the biological lower limit is; 0 fine but anything non-zero is a guess).

RudolfCardinal:
Having said that, the algorithm imposes a lower limit of 2 for CRP [ https://www.asas-group.org/clinical-instruments/asdas-calculator/ and https://www.asas-group.org/app/download/11136827897/Quick_ASDAS-CRP_Calculation_Form.pdf?t=1501230910 ]. The ESR limit is 0 [ https://www.asas-group.org/app/download/11136827797/Quick_ASDAS-ESR_Calculation_Form.pdf?t=1501230910 ]. So we should probably use a lower limit of 0 for both data files but then use max(CRP, 2) for the algorithm.

martinburchell:
>  So we should probably use a lower limit of 0 for both data files but then use max(CRP, 2) for the algorithm.

@RudolfCardinal OK I'll make those changes to this PR

PR #14, opened 2019-07-25
Add Short-Form McGill Pain Questionnaire (SF-MPQ2)

martinburchell:
@RudolfCardinal I need a bit of guidance with documenting copyright info for the two PubMed refs you found (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5221718/ and https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3225325/). I see IPR, History and Source sections of docs/source/tasks/sfmpq2.rst but I don't know which of these sections is applicable.

Secret strings commit 499c9dc04620f50ca2b74934dd5c057abbc71dd9

Also includes code documentation from https://github.com/RudolfCardinal/camcops/pull/13

### Checklist for new tasks:

- [X] Run self_test on server


RudolfCardinal:
Looks great! Have added bits to the docs as we discussed, and also the client was providing NULL as means for incomplete question groups. Nothing wrong with that, but the server was providing values, so changed for consistency; providing a value is probably the better of the two as it gives more info from incomplete tasks.

PR #15, opened 2019-07-30
Disease Activity Score-28

martinburchell:
### Checklist for new tasks:

- [X] Run self_test on server


martinburchell:
@RudolfCardinal I've used different cut-offs for DAS28-CRP and DAS28-ESR as per https://rmdopen.bmj.com/content/3/1/e000382. Not sure if we want to do that.

I don't know what you think about the HTML table on the client task detail popup. Overkill? I found I had to use inline styles as editing main.css made no difference.

RudolfCardinal:
I like the client HTML table! Minor tweaks: (a) style trivia; (b) the server calculated its summaries twice (once for rounding, once again); (c) capturing clinician info explicitly ("observer" in this task's jargon) and associated page type; (d) string constants for a couple of fieldnames client-side; (e) "x" for "no" rather than blank (probably a bit clearer). Good to go. I agree with the use of the 2017 CRP criteria (added to .RST file).

PR #17, opened 2019-08-02
Add Snaith–Hamilton Pleasure Scale (SHAPS)

martinburchell:
### Checklist for new tasks:

- [X] Run self_test on server


RudolfCardinal:
Lovely! Have fixed the countWhere() problem as per other comment. The only substantive thing I've changed is that the task was marked as having a clinician, but it's a self-report measure (so removed a mixin, changed the C++ flag, removed those fields from the upgrade script).

PR #19, opened 2019-08-02
Add optional waist circumference to BMI task

martinburchell:
Branched off shaps-dev so best reviewed once https://github.com/RudolfCardinal/camcops/pull/17 is merged

I've not tried to categorise the measurements as it appears to be quite a complex area. https://apps.who.int/iris/handle/10665/44583 might be a good starting point. 

I've not updated the SNOMED codes either. I'll need some guidance on that.

### Checklist

- [X] Run self_test on server


RudolfCardinal:
Thanks! Reviewed and will now merge. Have added the SNOMED coding (q.v.). Procedure is (1) establish codes/descriptions via e.g. https://termbrowser.nhs.uk/; (2) add to special XML with an arbitrary but meaningful tag; (3) add tags to cc_snomed.py; (4) build a SNOMED expression in the task. This is moderately complex and not everybody uses the grammatical nature of SNOMED (probably many ways to do something!) so I can't guarantee I have this optimal. Would be good to chat through in person non-urgently at some point (the SNOMED aspects have never yet been urgent!).

PR #21, opened 2019-08-22
QuDateTime: Add setMinimumDate() and setMaximumDate()

martinburchell:
Also fixes the default minimum of 1st Jan 1880 which wasn't having any effect. See https://bugreports.qt.io/browse/QTBUG-77775


RudolfCardinal:
Thanks! Only tweak was const-ref parameters. I then added some test code (widget test menu, demo questionnaire) and found a bug in the "set to now" button when using a time-only view, so fixed that.

PR #33, opened 2019-09-17
Debugging code for crash in QMediaPlayer() ?via gstreamer

RudolfCardinal:
Martin -- if you run the 3D ID/ID task, do you get a crash (or v. long pause) in soundfunc.cpp, when it calls QMediaPlayer()? A stack dump is in the comments in that file. I think this might be machine-specific (and has possibly gone wrong since an OS upgrade); looks like it relates to gstreamer. Am a bit stuck so if you have any insights that'd be great :) This branch just adds some warning code so the cause is more apparent when it goes wrong. It looks like it calls into video code; I don't know if it's possible to say "this media player is for sound only", which might conceivably help...?


martinburchell:
No crash on my desktop or laptop. With Qt in debug I couldn't see any more information. I guess we'd need a debug build of GStreamer.

Anyway the debugging code is useful so we might as well merge this

PR #34, opened 2019-09-19
Don't try to reindex during migrations

martinburchell:
@RudolfCardinal I'm not sure if you were thinking the warning should go in migration 0013 in place of the call to `reindex_everything` or once the database upgrade was complete. I've opted for the latter as I think the message will otherwise be lost in the noise.

### DIFF
--- a/server/camcops_server/alembic/versions/0013_task_index.py
+++ b/server/camcops_server/alembic/versions/0013_task_index.py
@@ -114,13 +114,6 @@ def upgrade():
 
     # ### end Alembic commands ###
 
-    if not context.is_offline_mode():
-        # Offline mode means "print SQL only". So we only execute the following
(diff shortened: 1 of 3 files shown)


martinburchell:
> Do you think we should go further and automatically re-index by default (or unless a "--no_reindex" flag is passed)?

That did occur to me. Would there be problems if someone downgraded/upgraded below 0013? I doubt anyone would do that.

martinburchell:
@RudolfCardinal I've added code to reindex when updating to head (unless the `no_reindex` parameter is passed in). If we try to reindex when downgrading or upgrading to a revision other than head there might be problems (eg if the `waist_cm` column on the bmi task table exists on the model but not in the database). So maybe we leave that to the discretion of the user.

PR #36, opened 2019-09-19
Fix undefined extra_msg in password templates

martinburchell:
`extra_msg` is not defined in `change_other_password.mako`. It is passed into
`change_own_password.mako` (as empty string) but isn't used. Assuming it is OK
to remove this in both places.


RudolfCardinal:
Yup, looks redundant; agree.

PR #40, opened 2019-09-27
APEQ CPFT Perinatal Report

martinburchell:
First iteration on APEQ CPFT Perinatal Report.

I'll put the date range filter in a separate PR.
TSV export is there as an option but gives the same as HTML. Once we've heard back from CPFT I'll either remove the option or implement a spreadsheet export as we discussed.

With the `DemoDatabaseTestCase` refactoring is there a way to type-hint the return values of the two `create_patient` methods without introducing circular imports?

### DIFF
--- a/server/camcops_server/alembic/versions/0013_task_index.py
+++ b/server/camcops_server/alembic/versions/0013_task_index.py
@@ -38,11 +38,10 @@ Creation date: 2018-11-10 22:54:18.529528
 # Imports
 # =============================================================================
 
-from alembic import context, op
+from alembic import op
 import sqlalchemy as sa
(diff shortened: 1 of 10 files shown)


RudolfCardinal:
Great! Have done the type hints (and a type-checking run).

Re circular imports: usual method is

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:

    from module_importing_us import Blah


def something():

    x = blah  # type: Blah

    ...


def other(q: "Blah"):

    pass
```

I think this is improved in [Python 3.7, PEP563](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) but it works fine like this.

PR #43, opened 2019-10-04
MAAS: Fix missing XML string

martinburchell:


RudolfCardinal:
Good catch -- thanks! Could you please also change the content for q2_a4 to "Fairly strong" (also wrong! Currently a duplicate of q2_a3. Sorry), then merge? This one should go in the changelog too as it's a functional change. Thanks!

PR #57, opened 2019-11-11
Data dump by email

martinburchell:
Adds option to have basic research data dump sent by email (via celery backend).

As part of this change I've moved some of the server email configuration out of the export recipient section and into the site section. This has meant some refactoring of the config constants into `cc_constants.py` and the removal of a duplicate `ConfigParamExportRecipient` class.

I've added serializers for `TaskCollection`, `TaskFilter `and `CamcopsRequest` though currently they encode the bare minimum data needed for the collection of tasks to be reconstructed.


RudolfCardinal:
Lovely! Have extended it to SQLite/SQL dumps and added the "queue for download" option.

PR #58, opened 2019-11-15
Continuous integration setup with Travis

martinburchell:
I've set up Travis CI to run various tests on every GitHub push:
* Install python packages
* Check package dependency problems
* Check packages for vulnerabilities
* Style and python errors (flake8) - same as the pre-commit hook we already have
* Run the automated tests

I've developed this on my own CamCOPS fork as you need to be the owner of the repository in order to set up Travis (if you're OK with it, you need to allow Travis to access your GitHub repository by going to https://travis-ci.org/RudolfCardinal/camcops/)

You can see the output of a build here:

https://travis-ci.org/martinburchell/camcops/builds/612366868?utm_medium=notification&utm_source=email


martinburchell:
I should add that I've fixed a few things along the way, including upgrading numpy

RudolfCardinal:
Looks fine; thanks! Only thing is a typing check: that "def _self_test" and "def self_test" promise to return None and don't (maybe remove "return" from the former, and add a return type to the latter)? Are you happy to fix/merge?

RudolfCardinal:
Sorry, see it perhaps needs to return something; just needs the return type to match.

martinburchell:
@RudolfCardinal should be good to go now

PR #61, opened 2019-12-10
Script to populate the database with random-ish patients and tasks

martinburchell:
This script will create five patients and one instance of every task per run. It will try to use valid values for integers and floats if limits have been set. Text fields are filled with nonsense. Uses faker for plausible names, dates of birth etc.

There's a bit of duplicate code from cc_unittest.py. At some point we might want to look at one of the test fixture libraries (eg factory_boy) for testing.

There's obvious scope for improvement though this has been good enough for testing REDCap.

### DIFF
--- a/server/camcops_server/camcops_server.py
+++ b/server/camcops_server/camcops_server.py
@@ -186,6 +186,13 @@ def _downgrade_database_to_revision(
         confirm_downgrade_db=confirm_downgrade_db)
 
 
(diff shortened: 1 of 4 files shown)


RudolfCardinal:
Nice! Hadn't heard of faker; looks good. I've made some trivial edits for PyCharm ("Instance attribute ... defined outside __init__") and reset the master RNG after re-seeding it.

PR #63, opened 2019-12-17
Export tasks to REDCap

martinburchell:
@RudolfCardinal Since I last showed you this, the main change is we're now querying REDCap for the next record ID / next instance ID

It also seemed neater to have all of the tasks/instruments in one fieldmap rather than one fieldmap per task so I've changed that.


RudolfCardinal:
Brilliant! As discussed re anonymous tasks/testing. A few docstring/minor doc changes pushed.

martinburchell:
I've tested with TASKS empty in the config (ie all tasks) and these are handled gracefully if an instrument is missing from the fieldmap.

PR #64, opened 2020-01-09
Fix alembic type comparison when creating migrations

martinburchell:
This should stop all those spurious TINYINT / Boolean changes appearing in migration files

I don't know why the previous approach stopped working. This now seems to be the recommended way to solve the problem.


RudolfCardinal:
Thanks! That's great.

PR #65, opened 2020-01-23
SHAPS: Document scale variation in studies

martinburchell:
Joel pointed me to a bunch of other examples, but maybe just one will do!:

https://www.ncbi.nlm.nih.gov/pubmed/30107636
https://www.ncbi.nlm.nih.gov/pubmed/25864641
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2957191/


RudolfCardinal:
Thanks. I'd include them all, with the others just as a bullet-point list. Good then to merge.

PR #67, opened 2020-02-03
Bugfix re generating PatientIdNum lineage records; generic code for d…

RudolfCardinal:
…oing this across all database objects


martinburchell:
@RudolfCardinal In order to understand the expected behaviour better, I've added a test which fails on the current master branch and passes now on this one. I'm not sure it describes a realistic scenario, though. Perhaps you could take a look at https://github.com/RudolfCardinal/camcops/pull/67/commits/97062ad3bee88340f85aa233c5111ed8a3b5d5c6

PR #68, opened 2020-02-07
Bugfixes for ID policies that prohibit direct identifiers

RudolfCardinal:
Patients with no DOB wouldn't upload correctly (JSON problem).
Error in policy checking on client side re "otherdetails" (versus "other" as the fieldname).
Added button to nullify DOB for the "no-DOB" policies (otherwise impossible!).


martinburchell:
Looks good!

Issue #82, opened 2020-06-01
Package vulnerabilities


martinburchell:
https://travis-ci.org/github/RudolfCardinal/camcops/jobs/693364730

reports the following vulnerabilities:

```
| package                    | installed | affected                 | ID       |

+============================+===========+==========================+==========+

| sphinx                     | 1.7.6     | <3.0.4                   | 38330    |

| pipenv                     | 2018.11.2 | <2020.5.28               | 38334    |
```

I have no idea where pipenv is coming from. pipdeptree reveals nothing.

@RudolfCardinal I notice we're using an old version of Sphinx. I'm going to try version 3. Are you are aware of any problems with newer versions?

martinburchell:
FWIW the description in the security database says "Sphinx 3.0.4 updates jQuery version from 3.4.1 to 3.5.1 for security reasons."

PR #83, opened 2020-06-02
Upgrade Sphinx, replace developers' paths with examples in documentation, remove pipenv installed by Travis

martinburchell:
Fixes https://github.com/RudolfCardinal/camcops/issues/82

Travis-CI reported two packages with vulnerabilities 

```
| package                    | installed | affected                 | ID       |

+============================+===========+==========================+==========+

| sphinx                     | 1.7.6     | <3.0.4                   | 38330    |

| pipenv                     | 2018.11.2 | <2020.5.28               | 38334    |
```

pipenv is installed by default by Travis-CI and as we don't need it, we can just uninstall it before running our scripts

I've upgrade Sphinx to 3.0.4 with no obvious problems. I've fixed a call to the deprecated `app.add_stylesheet`

I've also taken the opportunity to remove the paths in the generated config files in the documentation so they no longer refer to developers' own environments and won't change every time a different developer updates the documentation. I was able to do this mostly by checking an environment variable. The exception is the "Usage:" line from the client help file where I had to post-process the output. Qt doesn't make this easy to override.

I've left the timestamps and secret key for now so those are still going to change when the documentation is updated.


martinburchell:
> Looks good! Suggest also removing timestamp from autogenerated Apache and supervisord files (thought we'd done that, for some reason), e.g. "Created by CamCOPS server version 2.3.7 at 2020-06-01T16:34:46.822277+01:00." -- leads to unnecessary Git changes when rebuilding the "inclusion" docs. Thank you!

@RudolfCardinal Just to be clear, are you suggesting we remove the timestamp altogether from these files, or just in the examples in the documentation?

RudolfCardinal:
I meant removing the code that time-stamps the autogenerated files (leaving the version, though, which should deal with major changes). Does that seem reasonable?

martinburchell:
> I meant removing the code that time-stamps the autogenerated files (leaving the version, though, which should deal with major changes). Does that seem reasonable?

Yes I don't think the timestamps add much value. I'll do that then.

PR #86, opened 2020-06-30
Legal: (1) explicit mention of cookies; (2) re-ack. when terms change, (3) code ready to replace copyright messages upon IP transfer. Also (4) offer column info in downloads, (5) set language for new users.

RudolfCardinal:


martinburchell:
@RudolfCardinal I think this needs master branch merging into it. I don't understand why, because it looks like you have already merged the docker branch, yet there are still changes from that branch showing under "Files changed"

martinburchell:
Looks good! Let's hope the IP transfer happens before 2021.

PR #88, opened 2020-08-11
change_other_password bug fix

mtm93:
Fixed 'Missing user for id' error when changing own password through change_other_password. 

The GET request in change_other_password was calling the change_own_password function. This meant that because the page URL was unchanged, the POST request would still be on the change_other_password view. The hidden user_id html element is not required for the change_own_password function, so the user_id check failed. 

https://github.com/mtm93/camcops/blob/change_other_password_fix/server/camcops_server/cc_modules/webview.py#L781

The solution was to add a redirect instead of calling the change_own_password function.


martinburchell:
@RudolfCardinal I helped Matt with this. We weren't quite sure why changing your own password should be handled as a special case. Can you remember? Another solution would be to ensure the form's action attribute was set.

PR #91, opened 2020-08-11
Single user mode and task scheduler

martinburchell:
Support for Task Schedules (on the server) and Single User Mode (on the client)  to allow patients to complete tasks at home on their own devices.

@RudolfCardinal ready for you to break!


server/camcops_server/cc_modules/cc_forms.py
```diff
@@ -983,20 +984,20 @@ def after_bind(self, node: SchemaNode, kw: Dict[str, Any]) -> None:
...
-    missing = None
-
-    def __init__(self, *args, **kwargs) -> None:
-        self.allow_none = True
-        super().__init__(*args, **kwargs)
-
-    @staticmethod
-    def schema_type() -> SchemaType:
-        return AllowNoneType(Integer())
+# class OptionalWhichIdNumSelector(MandatoryWhichIdNumSelector):
```
martinburchell, line 987: Do we need this class any more? Can we remove it altogether?

server/camcops_server/cc_modules/cc_forms.py
```diff
@@ -1016,22 +1017,24 @@ def after_bind(self, node: SchemaNode, kw: Dict[str, Any]) -> None:
...
-    Optional node to capture an ID number value.
-    """
-    default = None
-    missing = None
-
-    @staticmethod
-    def schema_type() -> SchemaType:
-        return AllowNoneType(Integer())
+# class OptionalIdNumValue(MandatoryIdNumValue):
+#     """
```
martinburchell, line 1021: Do we need this class? Can we remove it altogether?

tablet_qt/core/networkmanager.cpp
```diff
@@ -878,30 +878,65 @@ void NetworkManager::updateTaskSchedules()
...
+        );
+        statusMessage(message);
+        fail(ErrorCode::JsonParseError, message);
+        return;
+    }
+    const QJsonArray patients_json_array = patient_doc.array();
+    const QJsonObject patient_json = patients_json_array.first().toObject();
+    Patient* patient = m_app.selectedPatient();
+    if (patient) {
+        patient->setPatientDetailsFromJson(patient_json);
```
martinburchell, line 922: This now means that the "needs upload" icon now never disappears! Is that OK or should we not trigger "needs upload" here?

martinburchell (commented):
A few comments. I don't think any of them are show-stoppers. You'll see I've pushed a few more changes.

server/camcops_server/cc_modules/cc_forms.py
RudolfCardinal: Can remove!

server/camcops_server/cc_modules/cc_forms.py
RudolfCardinal: Can remove!

tablet_qt/core/networkmanager.cpp
RudolfCardinal: Ah, yes, that is confusing. Agree; we should suppress the "needs upload" icon if this happens and there are no tasks present, I suppose (it's not that data hasn't changed, but we/the users wouldn't care without any tasks). If there are tasks present on the client afterwards, though, things have changed and do need uploading (is that right?).

martinburchell: I've suppressed the icon when updating the task schedules.

I realise we can't really update the task schedules before upload because it will mess up the complete status of the tasks yet to be uploaded. We could create a separate API call to update a patient's details and do this prior to upload but still fetch the task schedule updates after upload

Issue #94, opened 2020-09-04
Adding a new group does not do the same checks on group name as when editing


martinburchell:
If you add a new group called "Test group" and then edit the group, the form submission will fail saying 
Invalid group name (must be between 1 and 191 characters and contain only alphanumeric, hyphen, or underscore characters).

We should apply the same validation to the group name when it is first created.

@mtm93 would you like to have a go at fixing this?

PR #95, opened 2020-09-06
Updating 'addGroup' form name validation

mtm93:
Updated the "addGroup" form group name validation to match the "editGroup" form https://github.com/RudolfCardinal/camcops/issues/94


martinburchell:
Oh that was easy. Thanks!

PR #100, opened 2020-10-29
RAPID3 task

martinburchell:
### Checklist for new tasks:

- [X] Run self_test on server


RudolfCardinal:
I've merged in from task-scheduler-dev/basdai-dev (and fixed a database script problem that Alembic sometimes generates, where it converts a generic string field to a MySQL-specific thing when you read back from a MySQL database), but I can't see a rapid3.rst so don't have the original task reference to check against; would you mind adding that, please? Apologies if I've missed it!

martinburchell:
I'll aim to do the documentation today and then I'll move this PR out of draft status.

martinburchell:
@RudolfCardinal now documented.

A couple of points:

I've used https://www.rheumatology.org/Portals/0/Files/RAPID3%20Form.pdf as a template, which has 0.5 granularity on the scales. The 2009 paper doesn't include the questions but refers to the 1999 MDHAQ, which RAPID3 i based on. This includes "mark the line" scales.

There seems to be some variation as to whether to use the raw 0-30 RAPID3 score as the final measure or whether to divide by 3 to get a 0-10 score (for comparison with other RAPID indices). At the moment, we're keeping the score 0-30.

Any thoughts?

RudolfCardinal:
Looks good! Trivial tweaks only (mainly explanatory notes about the slightly odd scoring system).

PR #102, opened 2020-11-23
Set a default for Celery worker maxtasksperchild in the demo config

### LINKED ISSUE #96: The demo config should set a default for --maxtasksperchild in CELERY_WORKER_EXTRA_ARGS
To avoid the memory leaks experienced by the Copenhagen team we should limit maxtasksperchild in the demo config. Let's start with 1000 and drop it again if necessary

https://camcops.readthedocs.io/en/latest/administrator/resource_usage.html?highlight=CELERY_WORKER_EXTRA_ARGS#memory
https://camcops.readthedocs.io/en/latest/administrator/server_config_file.html#celery-worker-extra-args

@mtm93 another one for you if you have time

martinburchell:
Fixes #96


RudolfCardinal:
Looks fine!

PR #117, opened 2021-01-26
iOS version of client app

martinburchell:
Contains necessary changes to Qt project file and build script to build and deploy the app with macOS Catalina, QtCreator and XCode 12 using a development provisioning profile. There may be further changes needed when we submit to the Apple Store.

Some other changes were necessary to get the app looking decent on devices. Dialogs are full screen on iOS so many of our custom dialog classes were appearing stretched. There is also no close button on iOS windows so I've added a cancel button to those windows that were missing one.

I've also turned off some annoying things like autocapitalisation and predictive text in places and made the patient access key entry widget auto-insert the hyphens.


RudolfCardinal:
Looks good! I don't have the iOS kit to test properly, though. I've added a "-Wno-deprecated-copy" for clang as well as gcc (in camcops.pro) as I seem to be using that now and it wasn't compiling.

martinburchell:
> Looks good! I don't have the iOS kit to test properly, though. I've added a "-Wno-deprecated-copy" for clang as well as gcc (in camcops.pro) as I seem to be using that now and it wasn't compiling.

Hmm... Interesting. I thought I had fixed that.

We should get a tool to nag about / correct the code style like we do with the server code. I'd say it's consistency over obsessive-compulsion.

martinburchell:
> Is that because the language has made it illegal (so it's a version thing rather than a platform thing), or just duff compatibility?

I'm not really sure what is going on. All I can tell is:

* Apple Clang 12 doesn't support no-deprecated-copy so it won't compile on iOS / simulator with these set
* When parsing the camcops.pro file, both 'gcc' and 'clang' are set for iOS so to get this ignored for iOS we have to change these to linux-gcc and linux-clang . Alternatively we could use !ios.

Does this now compile on your Ubuntu?

RudolfCardinal:
Unfortunately not! This works: "gcc | clang". This works: "gcc | linux-clang". This doesn't: "linux-gcc | clang". So perhaps it's the "linux-" prefix not working. Mine is using GCC for compilation and clang for the Qt code editor. Is there a list of project file flags somewhere? I always struggle to find them. I have used a combination of "gcc | clang" and "!ios" (a nested "if" -- I haven't found a more sophisticated boolean syntax). Does this work on yours?

martinburchell:
> I have used a combination of "gcc | clang" and "!ios" (a nested "if" -- I haven't found a more sophisticated boolean syntax). Does this work on yours?

Yes!

PR #133, opened 2021-02-23
Privacy policy 2021-02-23

RudolfCardinal:


martinburchell (approved):
Looks good and very clear. I particularly like the fact that you've spelled out what we don't do.

PR #138, opened 2021-03-12
Schedule email templates

martinburchell:
Task schedules now support an optional email subject and body template with python string.Template style substitutions $access_key and $server_url. This should avoid introducing security holes.

I realise this is inconsistent syntax to the templates for exporting tasks as email. These use standard python string.format() with curly braces. It is possible to subclass string.Template to use a different search/replace regex but that seemed a bit complicated.

I've replaced the email column in the list of patients with a per-schedule email link.

Currently there is no form validation of the template. We could possibly check for invalid substitutions. Also some older browsers may have a maximum mailto: URL length.


RudolfCardinal:
Sounds good. I think ``${parameter}`` is also acceptable for ``string.Template`` (https://docs.python.org/3/library/string.html#template-strings).

(1) The alternative would be ``{parameter}`` and then substitute as per ``def get_export_filename(...)`` etc., with template_string.format(**d) and catching exceptions (previous code catches ``FORMAT_FAIL_EXCEPTIONS = (IndexError, KeyError)``). Is that any less secure?

One difference might be on failure. I think that ``string.Template`` will fail with "Your URL is ${wrong_parameter}", whereas ``str.format`` raises an exception so you would provide a default like "E-mail template malformed". Both are useless to the recipient; would the latter be easier to spot (or, for that matter, to pre-check for validity)?

(2) I realise also that the config file help never explicitly says "use braces", but just gives examples! I've fixed this -- and added ``ValueError`` (for unmatched braces) to ``FORMAT_FAIL_EXCEPTIONS``, now renamed ``STR_FORMAT_EXCEPTIONS``. I did this on master (sorry, that was wrong).

martinburchell:
`format()` will allow a template like:

`"{server_url} {access_key} {server_url.__class__}"`

and probably worse. It think this is less of an issue in the config file where the templates are known. The flexibility might be desirable in this case. `string.Template` doesn't allow this.

I'm currently using `Template.safe_substitute()` which won't raise an error if the template is malformed or a parameter is wrong. We could use `Template.substitute()` in the form validation which would catch these errors.

The thing that irks me the most about my solution is having two different substitution syntaxes at play. We could fix this by overriding the Template regex or using Template in both cases.

RudolfCardinal:
Ah. Good point. I see an illustration at https://lucumr.pocoo.org/2016/12/29/careful-with-str-format/ (and https://realpython.com/python-string-formatting/), though I think ``str`` objects don't "leak" quite this much information. I don't think the ``string.Template`` system will allow pure braces via its intended overrides; it seems to want a single character delimiter like ``$`` (optionally doubled) and then allows optional braces (hard-coded) after that.

A way to keep the current brace style (thus not breaking existing config files, plus it is quite legible) would be the sort of approach at https://stackoverflow.com/questions/15356649/can-pythons-string-format-be-made-safe-for-untrusted-format-strings, e.g.

    from string import Formatter
    
    
    class SafeFormatter(Formatter):
        """
        Allow brace substitution like ``str.format()``, but preventing "deep"
        access. See
        https://stackoverflow.com/questions/15356649/can-pythons-string-format-be-made-safe-for-untrusted-format-strings

        Examples:
        
        .. code-block:: python
        
            f = SafeFormatter()
    
            # Just like str.format():
            s1 = f.format("a = {a}, b = {b}", a=1, b="hello")  # OK
            s2 = f.format("{}, ID {}", a=1, b="hello")  # IndexError
            s3 = f.format("{a", a=1)  # ValueError
            s4 = f.format("{x}", a=1)  # KeyError
            
            # Things that will fail here:
            s5 = f.format("{x.__class__}", x=1)  # ValueError (new)
                    
            # ... but would not fail with str.format():
            t5 = "{x.__class__}".format(x=1)  # "<class 'int'>"
    
        """  # noqa
        def get_field(self, field_name, args, kwargs):
            if '.' in field_name or '[' in field_name:
                raise ValueError('Invalid format string.')
            return super().get_field(field_name, args, kwargs)

What do you think?

martinburchell:
@RudolfCardinal could you take another look?

I've created a SafeFormatter (in its own `cc_formatter` module - `cc_string.py` seems to be specific to XML strings).

The formatter takes an allowed list of keys rather than disallowing . So it's future-proof and avoids using any undocumented features of Python used in some examples I've seen online.

For now, the filename / email export continues to use the default Formatter.

RudolfCardinal:
Looks great! Good thinking.
I've added some minor docstrings.

PR #139, opened 2021-03-12
Changes necessary to deploy to the Apple Store

martinburchell:
In summary:

* Static instead of dynamic linking with OpenSSL
* Icons in the newer structure
* Correct version strings
* Bug fix when validating libraries in the Qt build script


RudolfCardinal (approved):
Terrific! Thank you. Are the other platforms happy with static OpenSSL linkage too?

martinburchell:
@RudolfCardinal I think it's just Android where we're linking dynamically now.

PR #140, opened 2021-03-16
MoCA now not permitted in the way it was before -- task effectively disabled (though will not delete old data)

RudolfCardinal:
See docs.


martinburchell:
Looks good from here. Did you want to note anything in the changelog?

RudolfCardinal:
Ah yes. Thanks!

PR #142, opened 2021-03-19
Release new version

martinburchell:
* Fix Qt static build problems in Linux and Windows
* Remove the Tornado safety check exception (CVE has been revoked)
* Bump all the versions
* Work in Progress on a release script


martinburchell:
@RudolfCardinal I'm going to just merge this as I think it's fairly non-contentious.

PR #149, opened 2021-03-30
Simplify manifest in

RudolfCardinal:


martinburchell (approved):
Looks much simpler. I'm pretty sure we don't need requirements.txt but if we find we do need one I see that according to https://stackoverflow.com/questions/43658870/requirements-txt-vs-setup-py there's a trick to route it through setup.py with just this one line:
`
 -e .
`

I don't think dependabot will cope with that. Dependabot ought to cope with setup.py but for some reason it doesn't like ours (or the one in CRATE).

Issue #145, opened 2021-03-26
Unresolvable cycles warning when creating a new alembic migration


martinburchell:
```
/home/martinb/workspace/camcops/working/.server_ve/lib/python3.6/site-packages/sqlalchemy/sql/ddl.py:1217: SAWarning: Cannot correctly sort tables; there are unresolvable cycles between tables "_security_devices, _security_users, patient", which is usually caused by mutually dependent foreign key constraints.  Foreign key constraints involving these tables will not be considered; this warning may raise an error in a future release.
  % (", ".join(sorted(t.fullname for t in err.cycles)),)

```

PR #154, opened 2021-04-13
Replace our test runner with pytest

martinburchell:
I've replaced our own test runner with pytest. This has the advantage of all the options pytest provides, such as:

* Automatic test discovery (all files that end in `_tests.py`)
* `-x` stop on failure
* `-k` test only according to wildcard (class name, file)

In addition the test database is reused if it already exists (unless the `--create-db` option is used) and each test is run inside a nested DB session, which is rolled back at the end of the test. So the database will be created at most once per test run, speeding up execution time considerably. 

I've preserved the option to create the test database in memory (`--database-in-memory`) but on my laptop with a SSD this is no faster.

pytest prefers function-based tests and its own fixture mechanism over `unittest.TestCase` subclasses with `test_*` methods and `setUp()` / `tearDown()`. It still works perfectly fine with the latter so there is no need to change that.  

The entire test suite now takes 3 minutes to run, which is nearly 5 times quicker than before.


martinburchell:
Also fixes #145

RudolfCardinal:
Nice!

RudolfCardinal:
After switching back to master I have a leftover "camcops_test.sqlite". Should this be in the repository, or the .gitignore file? (I'm guessing the latter?)

martinburchell:
Hmm interesting. That should have been dealt with in 8579c7b9ba6050c7916ef2fa0a427b9098282b6c
Was there a merge conflict?

RudolfCardinal:
Ah, no, sorry, just me being dim (switching to master but failing to re-pull from Github). All sorted!

PR #164, opened 2021-05-04
Fixes for penetration testing (per report of 28 Apr 2021)

RudolfCardinal:
Fixes for penetration testing (per report of 28 Apr 2021)


server/camcops_server/tools/run_server_self_tests.py
```diff
@@ -0,0 +1,63 @@
...
+
+from camcops_server.cc_modules.cc_baseconstants import CAMCOPS_SERVER_DIRECTORY
+
+
+DESCRIPTION = f"""- You can run tests manually via one of these methods:
+    pytest                                              # all tests
+    pytest FILE.py                                      # one test file
+    pytest FILE.py::some_function                       # one test function
+    pytest FILE.py::SomeClass                           # one test class
+    pytest FILE.py::SomeClass::some_member_function     # function within class
```
martinburchell, line 42:
```suggestion
    pytest FILE.py::SomeClass::some_member_function     # function within class
    pytest -k search_term                                                      # all test classes, functions etc matching 
```

martinburchell (changes requested):
Looks good! A couple of minor suggestions, which I can make.

We should get Travis (or maybe GitHub actions) pick up on more of these stylistic failings so you don't have to clean up after me all the time. Maybe have the docs autogenerate as well.

PR #168, opened 2021-05-11
Fix server errors when deleting task schedules

### LINKED ISSUE #150: Deleting a task schedule in use results in an internal server error
Probably a foreign key constraint not being handled cleanly

martinburchell:
Follows on from #169 

Fixes #150 internal server error when deleting task schedules with related items and patients.


RudolfCardinal:
I always struggle to get my head around these. So what follows is me thinking; sorry!

Focusing on the delete/cascade behaviour, I see (https://docs.sqlalchemy.org/en/14/orm/cascades.html#passive-deletes) that for a standard one-parent-many-children relationship, this is a slightly complex interrelationship between the Python deletion cascade and the database deletion cascade, and I presume the most important thing is that they don't overlap (will probably crash) and don't leave a gap (will leave undeleted records). In the example given, for "passive deletes":

* The (effectively) database-level command ``ondelete="CASCADE"`` is set on the child's foreign key (referencing the parent), and this means that when the database DELETEs a parent row, the database itself will delete linked children.
* The Python-level command ``cascade="all, delete"`` should be set on the parent (in its relationship to the child). The ``delete`` part means that when the parent object is Python-deleted, its children are marked for deletion (https://docs.sqlalchemy.org/en/14/orm/cascades.html#delete), and ``all`` is a synonym for a bunch of other things such as ``save-update, merge, ...`` (https://docs.sqlalchemy.org/en/14/orm/cascades.html).

In our existing code, we have this in, for example, web sessions:

    CREATE TABLE `_security_webviewer_sessions` (
    ...
    CONSTRAINT `fk__security_webviewer_sessions_user_id` FOREIGN KEY (`user_id`) REFERENCES `_security_users` (`id`) ON DELETE CASCADE
    ...

and that is an instruction that when the parent object is deleted (``_security_users``), the child object (``_security_webviewer_sessions``) should be automatically deleted by the database (cf. ``buildings`` and ``rooms`` in https://www.mysqltutorial.org/mysql-on-delete-cascade/).

As SQLAlchemy points out ("Notes on Passive Deletes") at that page, the database configures the cascade on the child object (the "many" side of a many-to-one relationship). The SQLAlchemy ORM configures it on the parent ("one") side.

None of our existing or new code uses ``passive_deletes``.

You key changes look to be that:

* when a ``TaskSchedule`` parent is deleted, Python now deletes its ``TaskScheduleItem`` and ``PatientTaskSchedule`` children -- via addition of ``cascade="all, delete"`` to ``TaskSchedule.items`` and ``TaskSchedule.patient_task_schedules``);
* when a ``Patient`` parent is deleted, Python now deletes its ``PatientTaskSchedule`` children (via addition of ``cascade="all, delete"`` to ``Patient.task_schedules``;
* when you delete a ``Patient`` parent or a ``TaskSchedule`` parent, the database no longer deletes its child ``PatientTaskSchedule`` (removal of ``ondelete="CASCADE"`` from the foreign keys specified in ``PatientTaskSchedule``); is that right?

On the ``TaskScheduleItem`` (child) / ``PatientTaskSchedule`` (parent) relationship, you don't currently have the child-based relationship have ``ondelete="CASCADE"``, and the ``passive_deletes=True`` flag isn't set on the parent. So is that the right method when you want SQLAlchemy to do all the work ("active deletes", I guess)?

I presume the tricky part is that ``PatientTaskSchedule`` effectively implements a many-to-many mapping between ``Patient`` and ``TaskSchedule``. The SQLAlchemy example (https://docs.sqlalchemy.org/en/14/orm/cascades.html#using-foreign-key-on-delete-with-many-to-many-relationships), which uses a raw ``Table`` rather than an ORM object for the mapping, uses ``passive_deletes`` for this.

Was the previous problem that both Python and the database were doing the deletion?

So -- nothing looks obviously wrong! I am not sure I understand this fully -- but presumably the key tests are:

(a) if a ``Patient`` is deleted, its ``PatientTaskSchedule`` entries are deleted without anything breaking;
(b) if a ``TaskSchedule`` is deleted, its ``TaskScheduleItem`` and ``PatientTaskSchedule`` entries are deleted without anything breaking;
(c) if the children are deleted, nothing deletes the parents.

Sorry in advance if this has not added to clarity!

martinburchell:
@RudolfCardinal Your analysis looks correct.

From memory, before this change, there was a foreign key constraint failure when trying to delete the task schedule. 

It seemed correct to enforce not-nullable on PatientTaskSchedule patient_pk and schedule_id
 
I tried various combinations of cascade behaviour. The final combination was the only one that didn't try to set non-nullable fields to null or leave orphaned rows.

Of your key tests, I'm testing (a) and (b) automatically as part of this PR. I think we're testing (c) already for Patients and PatientTaskSchedules. I've just tested it manually for TaskSchedules and TaskScheduleItems.

Issue #163, opened 2021-05-04
Put a signposting page at /api


martinburchell:
It's likely that a user is going to try launching the API URL used for patient registration with the server in a browser, especially if this appears as a hyperlink in an email.

We could route GET requests to /api to display a static page that says something like "You probably came here in error and you need to download and install the app etc"

RudolfCardinal:
Yes, good idea!

PR #177, opened 2021-05-27
Signposting page at /api for users who mistakenly put the address in a web browser

martinburchell:
Implements #163 

I'm sure we can do more here, such as link to the actual latest releases on GitHub (might be possible through the GitHub API).

It will need some Danish.


RudolfCardinal:
Looks good! Trivia:
- I think "Apple App Store" rather than "Apple Store"
- I've moved the URL constant to the constants file
- Worth continuing to Mako-filter the server URL -- I think that otherwise a silly and/or malicious user could enter garbage that's reflected (per the penetration testing). The GitHub URL is guaranteed (internally generated) but I am not sure we can guarantee that the server one won't contain hazards (if it includes query parameters).
- Danish done (by the machine, I hasten to add).

martinburchell:
@RudolfCardinal Filtering the line with the URL means the bold tags get escaped. Instead I'll recreate the `/api `URL from `req.route_url()`. That way we should avoid any garbage.

Issue #178, opened 2021-06-04
Bad 'filename' parameter: Invalid string. Permitted characters: A-Z, a-z, 0-9, '_'  when downloading file


martinburchell:
It is not possible to download a file from the download area:

This fails because it contains `-` and `.` characters: 

`/download_file?filename=CamCOPS_dump_2021-06-04T100622.zip`

RudolfCardinal:
Fix in https://github.com/RudolfCardinal/camcops/pull/179

PR #179, opened 2021-06-07
Fix validation of download filenames (validator was overzealous and blocking real content)

RudolfCardinal:
Fix validation of download filenames (validator was overzealous and blocking real content).

Fixes https://github.com/RudolfCardinal/camcops/issues/178.


martinburchell (approved):
Looks good. Thanks!

PR #181, opened 2021-06-07
DRAFT: Support for export to FHIR-based systems

martinburchell:
I've added FHIR export support to PHQ9 and APEQPT (anonymous). We could follow the same approach for other tasks.

The current approach has the task classes create the items for the Questionnaire and QuestionnaireResponse resources, with quite a bit of code duplication. We could make a further optimisation by putting a FHIR-like structure on each task and have some common task code read this to create the FHIR resources. BMI may be better described as FHIR Observation resource rather than QuestionnaireResponse. 

See the various TODOs throughout the code for other things we could add.

I'm not sure if I've interpreted the spec correctly for linking QuestionnaireResponse to Questionnaire with a URL.


RudolfCardinal:
I think I'm done fiddling -- see what you think! Of note: (1) the task list on the main menu, from which you can get to per-task views that show the FHIR Questionnaire structure in the abstract (most are autodiscovered -- imperfect but works, and all tasks also provide an HTML or XML or PDF DocumentReference); (2) there is a small-print FHIR option (next to the XML option) on each task, from which you can see an instantiated FHIR JSON view (though not tied to any particular recipient, and using slightly arbitrary PK-based patient identifiers as a result); (3) back-link URLs all work so you can go from the audit export views to the FHIR server and back. Also, pretty icons.

martinburchell:
Looks good and an improvement on what I did! Now we just need to try it on a real server.

PR #182, opened 2021-06-15
Minor test improvements

martinburchell:
Two new test case classes that remove the need to stub out create_tasks() for tests that don't require the "two of every task" database.

Also removes some noise and deprecated assertEquals() in the REDCap tests

Not urgent


RudolfCardinal:
Looks good!

PR #183, opened 2021-06-22
Support for emailing patient invitations from the server instead of the flaky mailto: links

### LINKED ISSUE #180: mailto: link not working on Windows
The long mailto:link with email subject and text body for the CPFT Perinatal team is not launching an email client on Chrome and IE on Windows.

A shorter mailto:test@example.com works fine

martinburchell:
For now I've stuck with plain text for the emails. I'll add a rich-text editor if I can get the automatic application launch on Android / iOS working. That will require hyperlinks with a camcops:// scheme and they will need to be HTML in order to appear navigable in an email.

This change also adds patient's forename and surname to the email template and removes some non-working unused test functionality to override config settings (probably broken since the move to pytest).

Fixes #180


RudolfCardinal:
Looks great. Question by email about missing access keys for patients created before the upgrade. With that answered, good to go.

martinburchell:
@RudolfCardinal Thanks! I think reply-to would be a good future addition to the schedule but not right now.

RudolfCardinal:
I think also some servers restrict "From" (so possibly From should be in a config file and Reply-To configurable by the user) -- not sure -- but definitely not a blocker for now!

PR #185, opened 2021-06-28
Launch app automatically from url on Android and iOS

martinburchell:
Follows on from #183 . Probably easier to review that one first.

Patients with Android phones/tablets or iPhones/iPads can now register themselves with the sever by launching the app from a hyperlink emailed to them. This bypasses the mode selection and patient registration dialog.

On Android the URLs are of the form http://camcops.org/register/ with query arguments for single user mode, server location and access key. On iOS the URLs are the same but we use the camcops:// scheme. This is the preferred approach on each platform. The disadvantage of the camcops:// scheme is that the URLs will not appear as hyperlinks in all mail clients, such as GMail, even when set as the href of HTML 'a' tags. The camcops:// links display correctly on the default mail client on my test iPad.

(At some point we might want to look at Universal Links, where you use http/https and have a website accompanying your app on the same domain. If the app isn't installed, your website can direct your users to the app store. If it is installed your users are taken straight to the app.)

If the CamCOPS app is not yet running, opening the link should start the app and use the defaults rather than display the mode selection dialog and patient registration dialog. If the app is running, but the user has not yet reached the patient registration dialog, opening the link should also use the defaults and skip registration. Currently there is no feedback to the user to indicate that this has worked, other than them not seeing these dialogs when starting up. We could change the new password dialog etc to listen for the incoming URL and display a message so the user knows that opening the link has worked.

Email templates and bodies are now created as HTML in the forms so that hyperlinks will be displayed correctly.
On the desktop you can automatically register a patient via command line arguments, though this is probably only of use during development.


RudolfCardinal (approved):
Looks good! I've made a couple of extremely trivial style tweaks. Not tested on Android/iOS but I've read through. The tab-separated command-line parameters surprised me a little but if it work, it works! Thank you! Great stuff.

PR #190, opened 2021-07-19
Assign UUIDs to patients that don't have them

martinburchell:


RudolfCardinal:
Looks absolutely fine logically to me. I am a bit wary about creating the extra stripped-down "Patient" class -- I have tested and pushed a version using SQLAlchemy Core. This makes zero practical difference so perhaps just a waste of time but see what you think (I guess the contrast is the way it was and revision 0061). The other thing I've done is check for mock connections -- both 0061 and 0066 were failing with the ``--show_sql_only`` flag because they required an actual query, so I've just skipped them in that context.

martinburchell:
@RudolfCardinal Happy to go with your method to avoid the duplicated class.

The approach I took for 0061 was from this StackOverflow post https://stackoverflow.com/questions/24612395/how-do-i-execute-inserts-and-updates-in-an-alembic-upgrade-script 

which I failed to credit. I copied this for 0066.

PR #191, opened 2021-07-20
Replace Travis CI with GitHub Actions for Continuous Integration testing

martinburchell:
I've replicated the exact same tests that we were doing with Travis CI to GitHub Actions only on Ubuntu 20.04.

Apart from different YAML syntax, the main difference is that with GitHub by default python packages are installed directly into the system installation (Travis gave us a clean virtualenv). So if we don't explicitly create our own virtualenv, things like `pip check` and  `safety check` break.

It would be better in the long term to split up the long "run" of commands into separate steps so it's obvious from the UI where the failure happened.

You can see the output in the Actions tab above (Python 3.9 builds still going as I type).

Also fixes the Python dependencies for the Debian package (would fail if 3.6 not present). This was not picked up with Travis because we were running on Bionic, which had python3.6 installed.

GitHub Actions looks promising for building Docker images and releasing to PyPI.


RudolfCardinal:
Terrific -- thank you! I've read the changes but have no expertise here so will leave this to you.

PR #193, opened 2021-07-29
CPFT Research Preferences task

martinburchell:
### Checklist for new tasks:

As part of this task, a MCQ can now have separate styles for each of the options. To do this I've changed the way NameValueOptions are shuffled. We keep the options in the original order but instead shuffle a list of indexes. This is so the link between the styles and the options is not lost. The confusing part is that the options can now be referred to by their _index_ or their _position_. For an unshuffled list these are the same thing. For a shuffled list the _index_ is the order in which the items are stored internally. The _position_ is their actual order when displayed.  Better suggestions for names welcome.

Awaiting information from David C. re IP.
Danish courtesy of Google Translate.
Also fixes yet another no-deprecated-copy failure, this time on linux / older gcc.


RudolfCardinal:
Looks great. Very trivial cosmetics only in the first commit. In the second, I think there were a couple of "external" users of NameValueOptions still using atIndex() -- I have made that protected to reduce the future risk (I can't see that we use that externally for other reasons) and switched them to atPosition(). That occurs in QuMcqGrid, QuMcqGridDouble, and QuMcqGridSingleBoolean. (Shuffling there would be odd, perhaps, but reversal certainly possible.) Is that correct?

martinburchell:
> Looks great. Very trivial cosmetics only in the first commit. In the second, I think there were a couple of "external" users of NameValueOptions still using atIndex() -- I have made that protected to reduce the future risk (I can't see that we use that externally for other reasons) and switched them to atPosition(). That occurs in QuMcqGrid, QuMcqGridDouble, and QuMcqGridSingleBoolean. (Shuffling there would be odd, perhaps, but reversal certainly possible.) Is that correct?

Yes that all makes sense. Unless you have any other comments and given we have the IP sorted I think this is OK to merge.

RudolfCardinal:
Thanks - no other comments!

PR #194, opened 2021-07-30
CPFT Covid Medical Questionnaire

martinburchell:
### Checklist for new tasks:

Follows on from #193. Please review that one first
"Questionnaire" is perhaps too grand a title as this is just one question.
Awaiting clarification on copyright from David C


RudolfCardinal:
Looks great. Let's check with DC: (1) "They were following an immunisation for COVID-19" -- should "were" read "started"? (2) Definition relates to when someone "caught" COVID-19 -- technically, they may well not know; should the question be phrased in terms of the onset of COVID-19 symptoms?

RudolfCardinal:
And one other question for D: at present, the context is implicit. Better to make it explicit? By which I mean: should we have an opening piece of text like: "In relation to the symptoms you are attending the post-COVID clinic about:" or similar? (At very first reading I thought it was talking about COVID symptoms themselves.)

PR #195, opened 2021-08-05
Any permitted user can manage and send emails to patients

martinburchell:
Two new user permissions to add/edit/delete patients on the server and send emails to them. Previously only a group administrator could do this.

The **Manage scheduled tasks for patients** menu option is now called **Manage patients and their tasks** and is no longer bundled with the Group Administrator options on the main menu.

Increased test coverage for user permissions.


RudolfCardinal:
Looks great! Just one question:

* In ``UserGroupPermissionsGroupAdminSchema``, the form defaults are ``True`` for the new options, but in the class ``UserGroupMembership`` the defaults are ``False``. Should these be consistent (and perhaps ``False`` as the more conservative setting)?

R

martinburchell:
> Looks great! Just one question:
> 
>     * In `UserGroupPermissionsGroupAdminSchema`, the form defaults are `True` for the new options, but in the class `UserGroupMembership` the defaults are `False`. Should these be consistent (and perhaps `False` as the more conservative setting)?
> 

@RudolfCardinal I see what you mean. What's odd is the form ignores these defaults (the checkboxes are all blank). I'll investigate.

martinburchell:
> @RudolfCardinal I see what you mean. What's odd is the form ignores these defaults (the checkboxes are all blank). I'll investigate.

Right, so the UserGroupMembership record already exists before we see this form so in fact the form will be populated from the database defaults rather than the form ones. The form defaults will be ignored.

PR #198, opened 2021-08-27
Multi-factor authentication

martinburchell:
Initial support for multi-factor authentication:

- TOTP using apps such as Google Authenticator
- HOTP (by email and SMS). 
 
SMS is via supported backends (currently Kapow and Twilio). It's easy to add more.

An admin can reset a user's MFA method but not . I've moved the advice on strong passwords from the Mako templates into the ZPT template for the widget. I've also updated all ZPT templates to not render the copyright block.

Currently the email address used for MFA is the existing one on the User class. So a user can now set or change this themselves.

A user can now have a phone number, which only they can change when editing their MFA preferences.

Possible improvements:
- Provide an alternative MFA method if (for example) the user has forgotten their phone
- The things in the checklists referenced at https://github.com/pyauth/pyotp

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -127,6 +127,8 @@ INSTALL_REQUIRES = [
 
     'twilio==6.63.0',  # SMS backend for Multi-factor authentication
 
+    'urllib3==1.26.7',  # dependency, pinned to avoid vulnerabilities
+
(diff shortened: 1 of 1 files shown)


martinburchell:
It all seems to be working as before. Thanks!

PR #213, opened 2021-10-07
Fix groups selector for non admin

martinburchell:
Fixes  https://github.com/RudolfCardinal/camcops/issues/211
Non admin group members with the "manage patients" privilege would see an empty group selector when adding/editing patients.


RudolfCardinal:
Looks good!

PR #220, opened 2021-11-19
Task schedule bug fixes and test NHS numbers page

### LINKED ISSUE #218: Adding a task schedule to a patient again wipes previous entries
Steps to reproduce:
* Create Schedule 1
* Assign Schedule 1 to a test patient.
* Do things (e.g. submit a task).
* Edit the patient, click "Add task schedule", and add Schedule 1 again.

Result:
It looks happy at the time, but the completed tasks vanish, and the patient is offered the tasks again -- perhaps a new schedule item is being created and displacing the old? (The old tasks are still there, of course, but their completion isn't recognized.)

(Currently testing in fhir-dev branch with v2.4.12 -- not finished/fully committed yet, so hang on before inspecting? But noting here.)

martinburchell:
Fixes #218. It should now be possible to add the same task schedule multiple times to a patient.
There is still a lack of association between a schedule and the completed tasks. If the time period of two schedules overlap, a completed task may end up appearing in both schedules. This is also true of a task that appears in two different overlapping schedules.

Simplifies the message displayed when task schedules are updated. The previous message was perhaps overly verbose.

Fixes a bug where deleting a PatientTaskSchedule would not cascade to PatientTaskScheduleEmail, resulting in an internal server error. This was because the delete was invoked from the Query rather than on the object directly.

Fixes the list of schedules in the selector on the EditPatient form to only display the ones the user can manage.

Adds a page of random NHS numbers for testing.


RudolfCardinal:
Looks great! I merged in https://github.com/RudolfCardinal/camcops/tree/fhir-dev, because that pymysql bug was stopping me testing, then made a few more visual tweaks (task schedule formatting and task details), so this is the cutting-edge branch now (sorry, that's a bit disorganized of me). Happy to merge when you're happy with the FHIR bits.

PR #221, opened 2021-11-22
Generate random data in task tests

RudolfCardinal:
(1) Make tests use the random data generation and ensure that they don't crash when rendering HTM.
(2) Fix the expectation/detection threshold task so it passes that test with random data.
(3) Fix bug where that recent style change in ``base_web.mako`` used ``request.matched_route.name`` when it might not exist (I think because ``request.matched_route`` was None but not absolutely sure) -- this caused a rendering crash to the log (though the PDF still got generated!). Not quite sure what ``base_web.mako`` is contributing to the PDF view, though.


server/camcops_server/templates/base/base_web.mako
```diff
@@ -38,7 +38,9 @@ from camcops_server.cc_modules.cc_pyramid import (
 %>
 
 <%block name="body_tags">
-class="route_${ request.matched_route.name }"
+    %if getattr(request, "matched_route"):  ## may be e.g. None in PDF rendering
```
martinburchell, line 41: Sorry! I actually didn't use this in the end so if it's more trouble than it's worth we could get rid of it.

RudolfCardinal: Generally I'd be indifferent but it's presumably a small performance hit on every page? If so, could just comment it out at this stage.

martinburchell:
Good idea!

I've removed the body class for now. It's there in version control if we ever want it back.

I mean to look at Factory Boy (https://factoryboy.readthedocs.io/) at some point for creating test fixtures. I've used it with Django and it claims to support SQLAlchemy.

Issue #228, opened 2022-02-25
Allow columns to be specified for basic spreadsheet dump


martinburchell:
The post-Covid Clinic would like to be able to specify which columns to include when exporting tasks to a spreadsheet

RudolfCardinal:
Pointlessly complex? Having too many columns is unlikely to be a significant burden. They can ignore ones they don't care about, the current system means they are guaranteed to get all relevant data, and sensible analytical strategies are not manual. Close and advise them accordingly, or is there some reason why this might be sensible?

RudolfCardinal:
On further thought: people wanting the simplified version aren't after GCP-type auditability (which remains there if they want) so version numbers out. Also, spreadsheet downloads offer a schema-style sheet (which adapts to being simplified). Also, the maximum version of numpy available for Python 3.7 might be lower than the minimum for the pointless security alerts, but one more try.

PR #229, opened 2022-02-27
Simplified spreadsheet export

RudolfCardinal:
Addresses https://github.com/RudolfCardinal/camcops/issues/228.


RudolfCardinal:
Failing on numpy safety checks. One affects "all version", though (">0") -- not fixable, therefore disable this safety check? Its concern is https://github.com/numpy/numpy/issues/19038. Adding "--ignore 44715" might fix this (not sure where it goes) -- could do that for a few others too, or move the minimum Python version to 3.8 as numpy 1.22.0 isn't available for 3.7.

martinburchell:
> Failing on numpy safety checks. One affects "all version", though (">0") -- not fixable, therefore disable this safety check? Its concern is [numpy/numpy#19038](https://github.com/numpy/numpy/issues/19038). Adding "--ignore 44715" might fix this (not sure where it goes) -- could do that for a few others too, or move the minimum Python version to 3.8 as numpy 1.22.0 isn't available for 3.7.

I'd do what I did for CRATE and stick with 3.7 and numpy 1.21. The ignores go in https://github.com/RudolfCardinal/camcops/blob/simplified_spreadsheet_export/.github/workflows/push-to-repository.yml#L59

We may find some of the warnings go away on Tuesday when the vulnerability database is updated.

RudolfCardinal:
Thanks; trying again...

martinburchell (commented):
It looks good from a code review perspective. I haven't tried it out yet. 

There's a dependency conflict in my virtualenv `kombu 5.2.3 has requirement amqp<6.0.0,>=5.0.9, but you have amqp 5.0.6.` and I need to check if that is just my stale environment. I'll look into the build failures also.

RudolfCardinal:
I have kombu==5.2.2 -- it's not in setup.py, so I guess we should pin that (but do feel free to bump everything to your higher versions!).

martinburchell:
> I have kombu==5.2.2 -- it's not in setup.py, so I guess we should pin that (but do feel free to bump everything to your higher versions!).

I rebuilt my virtualenv with the latest pip and now I don't have any broken dependencies

martinburchell:
Good to merge from my point of view. I did a quick test and it seems to be working as expected.

RudolfCardinal:
Thanks!

PR #235, opened 2022-04-11
Better task count report

RudolfCardinal:
Unifies the two "count task" reports, allowing arbitrary splitting by month/year/task type/user.
(Based off branch docs_web_user.)


martinburchell (approved):
Looks good

PR #241, opened 2022-07-05
EDE-Q task

martinburchell:
Initial implementation of the Eating Disorder Examination Questionnaire (EDE-Q 6.0) task.

I've also extracted out the height / weight / waist circumference question components from BMI to make them usable in this task and others. 

### Checklist for new tasks:


RudolfCardinal:
Looks v. good. I've made one small formatting change (one "sub-heading" or stem was in the same format as the major heading) and the other thing is that the max per subscale is 6, not 7 (and likewise the global score), I think. There's an outstanding question about handling optional questions if we're making some optional (affects .rst for documentation, .py for subscale calculation despite None values, .cpp); is that right?

martinburchell:
> Looks v. good. I've made one small formatting change (one "sub-heading" or stem was in the same format as the major heading) and the other thing is that the max per subscale is 6, not 7 (and likewise the global score), I think. There's an outstanding question about handling optional questions if we're making some optional (affects .rst for documentation, .py for subscale calculation despite None values, .cpp); is that right?

On this one Konstantinos said (on 27th May) "Please... make all of the questions mandatory and expect participants to answer them to the best of their knowledge. We don’t often see patients not being able to answer any of the EDEQ questions"

Good spot on the off-by-one error.

RudolfCardinal:
great -- I'll remove the "todo" bits three branches down the line!

PR #245, opened 2022-07-23
Initial development of CIA task

martinburchell:
### Checklist for new tasks:


RudolfCardinal:
Very nice. Only questions are (1) whether only some should be optional, rather than all (with attendant formatting questions), and (2) whether we should have "not applicable" out to the right of "a lot". Maybe one option is a grid but have some of them (4?) non-mandatory; if you tap, you've committed to answering, but if you don't, some will not show yellow and not block progress (but all this without an overt "not applicable" button, maybe with the " (if applicable)" text suffix as in the original for one of them)?

RudolfCardinal:
Mechanistically:
(1) In edeq.py, subscale() could call self.mean_fields(), which could have an additional option (e.g. accept_none, default True) to return None if one of the inputs is missing? I'm mainly thinking of the error handling, e.g. catching statistics.StatisticsError, that this function brings.
(2) In cia.cpp, since mandatory flags are set on the fields, I think a 4-option grid (values 0-3) would probably produce the "yellow = required" versus "white = not required" UI effect if a subset of answers is non-mandatory. Haven't tested it, though. And then the scoring would have to differentiate between "allowed missing" and not. Would this work visually? In the CIA I'm mainly thinking that the explicit "not applicable" radio buttons will confuse people and they'll pick the rightmost for "lots/bad" or the leftmost for "good" without careful inspection.

martinburchell:
>  In cia.cpp, since mandatory flags are set on the fields, I think a 4-option grid (values 0-3) would probably produce the "yellow = required" versus "white = not required" UI effect if a subset of answers is non-mandatory. Haven't tested it, though. And then the scoring would have to differentiate between "allowed missing" and not. Would this work visually? In the CIA I'm mainly thinking that the explicit "not applicable" radio buttons will confuse people and they'll pick the rightmost for "lots/bad" or the leftmost for "good" without careful inspection.

@RudolfCardinal  I've dropped the "not applicable" radio buttons and made q3, 4, 7 and 10 optional. To my eyes this seems to work well visually. What do you think?

RudolfCardinal:
looks good!

PR #244, opened 2022-07-18
Initial development of ISAAQ-ED appendix task and ISAAQ refactoring

martinburchell:
This follows on from https://github.com/ucam-department-of-psychiatry/camcops/pull/242. 

There's probably not much point in reviewing that one as this change extracts out the common ISAAQ functionality into base classes.

Either way, look at https://github.com/ucam-department-of-psychiatry/camcops/pull/241 first.

### Checklist for new tasks:

- [X] Add task to relevant menus
- [X] Generate code documentation
- [X] Update docs/source/changelog.rst
- [X] Test documentation rebuilds without errors
- [X] If the text of the task cannot be submitted here for copyright reasons, there is a template XML file under `server/camcops_server/extra_string_templates`.

Please [X] all the boxes above that apply

PR #246, opened 2022-08-02
Initial development of PARADISE-24 task

martinburchell:
Also contains #245, #244 and #241

I've not yet included the XML file with the full text. I'm assuming that given the measure is an attachment to https://doi.org/10.1371/journal.pone.0132410 which is published under a Creative Commons Attribution Licence,we can make it fully available.

## Checklist for new tasks:


RudolfCardinal:
Great; thank you! Yes, I agree - promoted to full availability. It is clearly stated to be open access and unrestricted use with attribution.

PR #252, opened 2022-09-26
Initial development of Installer

martinburchell:
Largely copied from CRATE. Unlike CRATE, the CamCOPS installer:

* Has an option to create a self-signed SSL certificate
* Can be set up to use a MySQL database outside of Docker
* Has an option to update the installation.

@RudolfCardinal I've added a `--password` option to `camcops_server make_superuser` so that the installer script will run on GitHub without user input. You may have deliberately chosen not to have this in the past. Another option would be to read the value from the environment variable. Django's `createsuperuser` supports both approaches.


RudolfCardinal (commented):
Thanks! Looks good but I've not managed to test yet -- as installer.sh hard-codes to release paths, I tried running installer.py within an existing venv, but python-on-whales isn't in setup.py. Is there a simpler test route? (Sorry, this is probably a daft question.)

martinburchell:
`installer.sh -d` to run from local checkout. This will bypass downloading the tar file

RudolfCardinal (approved):
Fantastic -- thanks!

Two minor queries:

1. Initially it builds/starts camcops_db, camcops_rabbitmq, then camcops_workers (which fails), then it creates the CamCOPS docker image. 

[+] Running 7/7
 ⠿ Network camcops_camcops_network         Created                                                                 0.5s
 ⠿ Volume "camcops_db_data_volume"         Created                                                                 0.1s
 ⠿ Volume "camcops_rabbitmq_volume"        Created                                                                 0.1s
 ⠿ Volume "camcops_camcops_config_volume"  Created                                                                 0.0s
 ⠿ Volume "camcops_camcops_tmp_volume"     Created                                                                 0.1s
 ⠿ Container camcops_db                    Created                                                                 2.1s
 ⠿ Container camcops_rabbitmq              Created                                                                 2.1s
[+] Running 2/2
 ⠿ Container camcops_db        Started                                                                             5.4s
 ⠿ Container camcops_rabbitmq  Started                                                                             5.4s
[+] Running 0/1
 ⠿ camcops_workers Error                                                                                           3.0s
...

... but then chugs on happily, so I presume that's OK. And didn't do it the second time round (once the image had been built?).

2. I got some errors because I already had $CAMCOPS_DOCKER_CONFIG_HOST_DIR set up with a config file (a MySQL connection error) -- is that condition worth detecting, do you think? The config file timestamp updated to today, so I think some editing happened, but the URL (with password) was unchanged so didn't match the one I'd typed in. It was happy the second time once I'd cleared out those config files.

martinburchell:
> Fantastic -- thanks!
> 
> Two minor queries:
> 
>     1. Initially it builds/starts camcops_db, camcops_rabbitmq, then camcops_workers (which fails), then it creates the CamCOPS docker image.
> 
> 
> [+] Running 7/7 ⠿ Network camcops_camcops_network Created 0.5s ⠿ Volume "camcops_db_data_volume" Created 0.1s ⠿ Volume "camcops_rabbitmq_volume" Created 0.1s ⠿ Volume "camcops_camcops_config_volume" Created 0.0s ⠿ Volume "camcops_camcops_tmp_volume" Created 0.1s ⠿ Container camcops_db Created 2.1s ⠿ Container camcops_rabbitmq Created 2.1s [+] Running 2/2 ⠿ Container camcops_db Started 5.4s ⠿ Container camcops_rabbitmq Started 5.4s [+] Running 0/1 ⠿ camcops_workers Error 3.0s ...
> 
> ... but then chugs on happily, so I presume that's OK. And didn't do it the second time round (once the image had been built?).

Yes. I don't know how to avoid that error message other than silence the output.
 
>     2. I got some errors because I already had $CAMCOPS_DOCKER_CONFIG_HOST_DIR set up with a config file (a MySQL connection error) -- is that condition worth detecting, do you think? The config file timestamp updated to today, so I think some editing happened, but the URL (with password) was unchanged so didn't match the one I'd typed in. It was happy the second time once I'd cleared out those config files.

I've changed it so the config file is always recreated by the installer but the old config file is retained under a new name in case there are things to copy across.

PR #253, opened 2022-09-29
Replace our pre-commit hook script with third party pre-commit package

martinburchell:
Commits are getting too slow with Black and PEP8 checks on every python file for every commit. We only want to check the files that are being committed .

[pre-commit](https://pre-commit.com/) will do this for us. For now I've just added a few checks, namely:

* Black
* Flake8
* YAML, JSON and XML syntax checking
* Stop large files being included

We might want to add more from the [complete list of supported hooks](https://pre-commit.com/hooks.html)

Usage:

Remove our pre-commit hook:

```
$ rm ~/workspace/camcops/.git/hooks/pre-commit
```

Inside the virtualenv:

```
$ cd ~/workspace/camcops/server
$ pip install -e .
$ pre-commit install
```

This will now run pre-commit on any committed files.

To run on all files:

```
$ pre-commit run --all-files
```

This change also ignores some PEP8 issues that were previously undetected.


RudolfCardinal (approved):
Looks good! Should something about "pre-commit install" go into the developer docs?

RudolfCardinal (approved):
[comment on wrong PR]

PR #255, opened 2022-11-08
Report to count server-created patients and their assigned tasks

martinburchell:
As requested by the Covid Clinic.

I've used [factory_boy](https://factoryboy.readthedocs.io) for creating test fixtures, which I've used before with Django and is well supported. It's a bit fiddly when a model class has its own __init__() I'll try refactoring our other test fixture code to use it at some point and see if it's worth it.

I've also upgraded to the latest SQLAlchemy 1.3, though I'll upgrade to 1.4 after I've done Qt6.2.


martinburchell:
> Thank you! Looks great. I thought it wasn't noticing new e-mails but turned out it was just my e-mail outbound system that wasn't working. (Of note: if the outbound system is broken, the front end does hang. Do you think we should push the "scheduled task e-mail send" to a back-end job? I think there is some potential that a misconfigured e-mail system gradually uses all of the front-end threads up.)

Yes https://github.com/ucam-department-of-psychiatry/camcops/issues/256
 
> Future note: I see that for SQLAlchemy, the ResultProxy object becomes CursorResult.

Noted

PR #260, opened 2022-11-30
Rich text for command-line help

RudolfCardinal:
Uses https://pypi.org/project/rich-argparse/. This seems pretty good; it adds colour to command-line help help, making it (I think) significantly more legible. Autodetection of being in a "pipe" environment so text versions are unchanged. Colour scheme designed to cope with light/dark colour schemes.


martinburchell:
Looks good! I don't know how this affects generation of documentation. See my comment in #259 re changes to formatting for Python >= 3.9.

The failing checks above will probably be fixed by merging #259 or cherry-picking the relevant bits.

PR #262, opened 2022-12-08
ACE-III update and Mini-ACE

RudolfCardinal:
### Checklist for new tasks:

Only failing test is because miniace.rst not published to camcops.readthedocs.io yet


RudolfCardinal:
There is also now ``tablet_qt/tools/clang_format_camcops.py``, which has destructive commands like ``modify`` (not yet used!) and information commands like ``diff`` (including ``--diffall``). This reads ``clang_format_camcops.yaml`` and attempts to apply a style reasonably similar to the Python/Black style. May need a bit more iteration; in particular it splits long lines like ``void somefunc();  // long long comment`` in between the brackets, though perhaps that just indicates a need for comment reformatting. Anyway, see what you think!

martinburchell (changes requested):
Looks good. A few minor things but I can fix those.

I've pushed a couple of changes to make GitHub Actions happy. I've changed the task documentation test so it just looks for the presence of the source .rst file in the docs folder. This means that we aren't dependent on an internet connection to run the tests.

I don't really have an opinion on code style apart from that we should have one and it isn't esoteric. [pre-commit has a hook for clang-format ](https://github.com/pre-commit/mirrors-clang-format) so we can check before committing (and get GitHub to nag us if pre-commit is not set up).

RudolfCardinal:
Thanks! I'm not sure if you're all done (I think so but a test is running) -- if so, feel free to merge, or I can if you prefer.

PR #267, opened 2023-02-10
Update client to Qt 6.2

martinburchell:
Definitely not ready for review yet!


martinburchell:
Need to change source branch to the qt6.5 one so closing...

PR #271, opened 2023-02-14
BMI: Make waist circumference optional again

### LINKED ISSUE #264: BMI: Waist circumference no longer optional
Since I separated the mass, height and waist circumference from the BMI task in v2.4.13, they are all now mandatory.

So the text says "optional" for waist circumference but the green square doesn't appear until you've filled it in. Workaround is to abort the questionnaire (it will be marked as "complete") or fill in a value for waist circumference.

Probably `QuWaist`, `QuMass` and `QuHeight` all need a mandatory flag (defaulting to true), which can be used in `setUpFields().` BMI can then make waist optional.

martinburchell:
Fixes #264 . QuMass, QuHeight and QuWaist can now all be specified as optional. For BMI we set the waist circumference field to be optional.


RudolfCardinal (approved):
Looks good, works well.

PR #272, opened 2023-02-14
Rename CAMCOPS_QT_BASE_DIR to CAMCOPS_QT5_BASE_DIR

martinburchell:
This makes things less painful when switching between Qt6 and Qt5


RudolfCardinal (approved):
Looks good. I think there was one more in build-qt.yml, which I've updated.

PR #274, opened 2023-02-17
Workflow fixes

martinburchell:
All the workflows should be passing again for now.

* Switch from PyPDF2 to pypdf, 
* Upgrade flower
* Remove py
* Split the pip-install-and-tests workflow into separate ones
* Add a new workflow to check for missing migrations
* Add some missing migrations and fix PHQ-8 and Mini-ACE migrations to workaround https://github.com/sqlalchemy/alembic/issues/326. I think it's OK to modify the original migration file here because these haven't yet appeared in a release.


RudolfCardinal (approved):
Thanks! Looks great. Sorry for missing the FKs.

PR #276, opened 2023-02-23
CHI-T task update

### LINKED ISSUE #269: CHI-T task update
Per correspondence from Konstantinos Ioannidis, Sam Chamberlain (e.g. 2023-02-03, 2023-02-13): CHI-T task update

This adds one item and (versus some earlier versions) removes a question about subjective impairment.

* a/w clarification re method (update/new/replace)
* cite new paper: https://pubmed.ncbi.nlm.nih.gov/36680457/
* a/w clarification re licensing

martinburchell:
I've updated CHI-T to add the extra response and remove Q16. Closes #269


RudolfCardinal:
Thanks; looks great.
* One local question by e-mail that you've already answered.
* I've added very slightly to the docs (``chit.rst``).

Thinking about potential version incompatibility:
* "Old server when trying to upload": we could set ``minimumServerVersion()`` in ``chit.h``/``chit.cpp``. The old task should upload fine structurally (the missing Q16 won't cause a problem) but the validators would later complain via ``field_contents_valid()``. I don't think adding this will cause much of a problem, so I've added it. (The task would also have missing strings if it fetched from an old server.) Say if you think that's wrong!
* "Old client, existing data": we think old client data doesn't exist anywhere for this task version, so can ignore this problem. (The server upgrade would have deleted data for Q16 if that weren't true. The consequences would be mis-labelling of the answers and a misleading total score, though your footnote plus the client version information displayed would allow the observant user to understand.) Agree to ignore.
* "Old client tries to upload": not sure if this is a problem. But I've set ``minimum_client_version`` in ``chit.py`` just in case. Reasonable? This will, via ``client_api.ensure_valid_table_name()``, make the client be upgraded before it can upload.

I also looked at trying to fix the clazy linter comments, e.g. about string literals, but some of the suggested performance fixes for Qt5 then instantly fall foul of warnings for Qt6. For example:

    const QString spacer = " ";  // bad: QString(const char*) being called [clazy-qstring-allocations]
    const QString spacer = QStringLiteral(" ");  // good

as per https://doc.qt.io/qt-5/qstring.html#more-efficient-string-construction and https://github.com/KDE/clazy/blob/master/docs/checks/README-qstring-allocations.md, but

    const QString suffix = "";  // bad: QString(const char*) being called [clazy-qstring-allocations]
    const QString suffix = QStringLiteral("");  // bad: Use an empty QLatin1String instead of an empty QStringLiteral [clazy-empty-qstringliteral]
    const QString suffix = QLatin1String("");  // bad: QLatin1Char or QLatin1String is being called [clazy-qt6-qlating1stringchar-to-u]

so presumably the best thing is to move directly to Qt6, e.g. this sort of framework: https://doc.qt.io/qt-6/qt-literals-stringliterals.html ?

martinburchell:
Your changes for version compatibility seem entirely reasonable to me.

> so presumably the best thing is to move directly to Qt6, e.g. this sort of framework: https://doc.qt.io/qt-6/qt-literals-stringliterals.html ?

That appears to be new in Qt6.4 which isn't LTS. I'm currently targeting 6.2, which is LTS.

PR #277, opened 2023-02-23
Replace ISAAQ with ISAAQ-10

martinburchell:
I'm pretty sure no one has recorded any ISAAQ tasks so I've opted to delete the table. It's possible ISAAQ may appear in a task schedule so I've included a data migration to change it. Less likely but possible is that ISAAQ appears in a task filter. If this is the case, I'll fix it manually at deployment time.  

### Checklist for new tasks:


RudolfCardinal (approved):
Looks good! One bug -- it failed to upload because of the old "isaaq" table on the client side. I have added ``Isaaq10::upgradeDatabase()``, called by ``CamcopsApp::upgradeDatabaseAfterTasksRegistered()``. This is conditional on a version change at present (presumably fine for end users but some risk of confusing us as developers). If you're happy, go for it.

Issue #278, opened 2023-02-28
"SSL handshake failed" when connecting to the CPFT server from iPad


martinburchell:
Seen on S3 ward iPad and also our development one (iPad 7, software version 14.6)

martinburchell:
Also fails on the University server

martinburchell:
Also fails with Qt 6.2.4 and OpenSSL 1.1.1s

PR #283, opened 2023-03-16
Use SecureTransport instead of OpenSSL when building Qt on iOS

### LINKED ISSUE #136: OpenSSL shouldn't be necessary with iOS
According to Samuel Gaist we should be able to use the iOS native framework instead of OpenSSL https://forum.qt.io/topic/121445/dynamically-linked-openssl-not-supported-on-iphonesimulator/7

martinburchell:
Closes #136, #278 

It's still a mystery why this is not working with OpenSSL.

https://developer.apple.com/documentation/security/secure_transport says "This API is considered legacy. Use the [Network](https://developer.apple.com/documentation/network) framework instead."

https://wiki.qt.io/Plans_for_Modules says "Investigate Apple's networking framework; SecureTransport was deprecated by Apple, need an alternative ". Qt are still shipping binaries with SecureTransport for iOS so I guess we wait to see what that alternative will be.

We still need to link OpenSSL when building CamCOPS so that SQLCipher will work.

### DIFF
--- a/tablet_qt/camcops.pro
+++ b/tablet_qt/camcops.pro
@@ -350,7 +350,7 @@ ios {
     STATIC_LIB_EXT = ".a"
     DYNAMIC_LIB_EXT = ".dylib"
     CAMCOPS_QT_LINKAGE = "static"
-    CAMCOPS_OPENSSL_LINKAGE = "none"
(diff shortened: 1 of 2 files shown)


RudolfCardinal (approved):
Looks good!

PR #285, opened 2023-03-28
Restrict memory used by celery workers

### LINKED ISSUE #280: Celery workers grow over time
The CPFT server is running out of memory over time. We suspect that something in the server code is leaking memory and the celery workers are hanging on to it. 

We already set `max-tasks-per-child` to 1000 but probably because the server is lightly used, the workers are not being replaced very often. As an experiment I've also set `max-memory-per-child` to 1,000,000 Kib so the total memory consumed by celery should never exceed 4Gb (1/4 of the memory available).

It would be good to identify where the memory is being leaked and an easier additional measure would be to set a sensible default for `max-memory-per-child` in the sample config file.

martinburchell:
These settings have helped reduce the memory used by Celery on the CPFT server. So I've added them to the demo supervisor and Celery configuration.
`max-memory-per-child` in the Celery args section of the CamCOPS config file set to a sensible default so no celery worker can grow beyond this size
`stopasgroup` in the supervisor config to stop orphaned processes being left behind when supervisor stops.
`startretries` in the supervisor config was already set to 10. Seems sensible. 

Fixes #280

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -3767,3 +3767,8 @@ Current C++/SQLite client, Python/SQLAlchemy server
 
 **Client and server v2.4.16, IN PROGRESS**
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Set default ``max-memory-per-child`` in the arguments passed to Celery workers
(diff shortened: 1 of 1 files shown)


RudolfCardinal (approved):
Looks great!

Issue #290, opened 2023-04-29
CherryPy failing to launch with local test certificates


RudolfCardinal:
With an old or a new self-signed certificate and ``camcops_server serve_cherrypy``, which used to work fine, after an upgrade to Ubuntu 22.04.2 LTS (but a clean CamCOPS venv installation), I'm getting:

    2023-04-29 16:05:12.756 [p1210581.t140646515499008] cherrypy.error:INFO: [29/Apr/2023:16:05:12] ENGINE Bus STARTING
    [29/Apr/2023:16:05:12] ENGINE Started monitor thread 'Autoreloader'.
    2023-04-29 16:05:12.756 [p1210581.t140646515499008] cherrypy.error:INFO: [29/Apr/2023:16:05:12] ENGINE Started monitor thread 'Autoreloader'.
    [29/Apr/2023:16:05:12] ENGINE Serving on https://127.0.0.1:8088
    2023-04-29 16:05:12.866 [p1210581.t140646515499008] cherrypy.error:INFO: [29/Apr/2023:16:05:12] ENGINE Serving on https://127.0.0.1:8088
    [29/Apr/2023:16:05:12] ENGINE Bus STARTED
    2023-04-29 16:05:12.866 [p1210581.t140646515499008] cherrypy.error:INFO: [29/Apr/2023:16:05:12] ENGINE Bus STARTED
    [29/Apr/2023:16:05:12] ENGINE Error in HTTPServer.serve
    Traceback (most recent call last):
      File "/home/rudolf/dev/venvs/camcops/lib/python3.10/site-packages/cheroot/server.py", line 1807, in serve
        self._connections.run(self.expiration_interval)
      File "/home/rudolf/dev/venvs/camcops/lib/python3.10/site-packages/cheroot/connections.py", line 198, in run
        self._run(expiration_interval)
      File "/home/rudolf/dev/venvs/camcops/lib/python3.10/site-packages/cheroot/connections.py", line 241, in _run
        new_conn = self._from_server_socket(self.server.socket)
      File "/home/rudolf/dev/venvs/camcops/lib/python3.10/site-packages/cheroot/connections.py", line 295, in _from_server_socket
        s, ssl_env = self.server.ssl_adapter.wrap(s)
      File "/home/rudolf/dev/venvs/camcops/lib/python3.10/site-packages/cheroot/ssl/builtin.py", line 270, in wrap
        s = self.context.wrap_socket(
[...]

PR #291, opened 2023-05-09
Workflow to test launching CherryPy with a self-signed certificate

martinburchell:
Originally set up to investigate #290, a new workflow to test launching the server with CherryPy with a self-signed certificate. The matrix of python/Ubuntu versions is overkill and we could reduce it to a single combination to save some energy.

In addition this PR fixes some problems with other workflows:

* The test MySQL user now needs to be created with `mysql_native_password`, otherwise MySQL assumes a different method and attempting to start the server fails with an obscure error from pymysql about cryptography not being present.
* Names of jobs were incorrect from copying and modifying yaml files


RudolfCardinal:
Looks good -- thank you! Maybe worth a safety check in ``.github/scripts/create_config_file.sh`` that ``CAMCOPS_CONFIG_FILE`` is set (non-blank), now that setting has been moved to another file?

martinburchell:
> Looks good -- thank you! Maybe worth a safety check in `.github/scripts/create_config_file.sh` that `CAMCOPS_CONFIG_FILE` is set (non-blank), now that setting has been moved to another file?

It's kind of what the -u switch is supposed to do, though in this case the script doesn't reach the variable until it has launched the server. So instead I'm dumping the value before running the command and this should fail quicker.

RudolfCardinal:
Sorry, my Bash ignorance -- that looks good.

Issue #126, opened 2021-02-05
EQ-5D-5L scale not visible on iPhone 8


martinburchell:
![Simulator Screen Shot - iPhone 8 - 2021-02-05 at 17 03 38](https://user-images.githubusercontent.com/192929/107069009-3d5cfb00-67d9-11eb-8ca3-fc56557cf6d0.png)

martinburchell:
Workaround is to put the number in the box

martinburchell:
There is a related problem reported by CPFT S3 ward with the scale not displaying in landscape on iPad.

Two possible solutions I can think of:

* Scale the scale to make it fit the height of the screen. Disadvantage is that some tasks may specify the dimensions of scales in centimetres so we'd lose some authenticity
* Add a scrollbar. Disadvantage is that the user might not think to scroll to see the lower values.

@RudolfCardinal any thoughts?

PR #294, opened 2023-06-09
Fix display of visual analogue scale on EQ-5D-5L task

martinburchell:
I've changed the QuThermometer scale factor in the EQ-5D-5L task to be calculated from the screen height. It won't auto-resize on orientation change (I've found it to be buggy in the past). I'll revisit this when we move to Qt6.

I've also added 20px (scaled) padding to the top and bottom images of the thermometer to stop clipping the top and bottom numbers. We might want to make this configurable in the future.


martinburchell:
Fixes #126

RudolfCardinal (approved):

PR #302, opened 2023-07-06
Update client to Qt6.5

martinburchell:
Definitely not ready for review!


RudolfCardinal (approved):
Looks excellent -- thank you! Camera working nicely under Linux. Small changes:

* trivial RST typography;
* I think a fix to GridLayoutHfw::getGeomInfo() -- was modifying a flag (gi.m_expanding) that was then overwritten from a local copy (expanding) -- **correct?**;
* removal of redundant header file inclusion (per new ?clangd warnings), with "// IWYU pragma: keep" for the preprocessor-related ones that it autodetects the need for poorly;
* local copy of scores in DAS-28 rather than x.toDouble() several times, and similarly for RAPID3;
* in passing, be explicit about CRP/ESR units for DAS-28;
* QuMass: fix warning about some non-slot functions being marked as slots;
* GAF: trivial modification of isComplete();
* fix some PyCharm warnings in build_qt.py

One other question: I'm getting some pixellation of icons (e.g. top right of main screen/questionnaires), which doesn't seem to be in the source PNG files. I've tried overriding physical DPI or not though (not surprisingly) that had no effect (in retrospect I think only for the sliders). Have you seen that, Martin?

RudolfCardinal:
Automatic C++ tests failing, though (I don't think I touched those!). I'm guessing ``tests/auto/db/field/testfield.cpp`` needs to ``#include "......./lib/version.h"`` so I've tried that. Now it passes though I can't see testfield.cpp in the output! Is it running, do you think? I'm not sure of the standard process for running the tests locally, though.

martinburchell:
> One other question: I'm getting some pixellation of icons (e.g. top right of main screen/questionnaires), which doesn't seem to be in the source PNG files. I've tried overriding physical DPI or not though (not surprisingly) that had no effect (in retrospect I think only for the sliders). Have you seen that, Martin?

No, do you have a screenshot?

PR #307, opened 2023-07-22
C++ Tests for Qt5

martinburchell:
Backported C++ tests from the Qt6.5 branch, partly to reduce the size of that pull request and partly to test the "before" state. The choice of tests is a bit random, covering some of the more easily testable code that has needed to change for Qt6.5.  The `NameValueOptions` tests are from a previously aborted attempt at using Qt Test. 

In order to reduce the number of dependencies needed to compile the code to be tested, I've split out some of the functions in `lib/uifunc`:

* Widget-related functions move to new `lib/widgetfunc`
* ` uifunc::escapeString() `moves to `lib/stringfunc`
* Some of the calls to `uifunc::stopApp()`, which are more checks for bad code and the user should not normally see are replaced with `errorfunc::fatalError()`. This aborts the app but doesn't pop up a dialog.
* The null pointer check in `soundfunc::makeMediaPlayer()` and  accompanying alert dialog move to the calling code.

We can probably do more to reduce the number of dependencies but this is enough to build the test executables quite quickly.

In the process of testing `convert::toSqlLiteral()` I've fixed two bugs with Date and Time handling. I'm not sure this code was being used anywhere.

There's a GitHub workflow for this on the Qt6.5 branch. You can't get qtmultimedia for 5.12.x with aqtinstall.


RudolfCardinal:
Having upgraded locally from Ubuntu 20 to 22, the old Qt 5 build doesn't run (perhaps not too surprising as would have been configured for the OS at the time). Rebuilding Qt5 now, however, defaults to "gcc" = gcc-11, but this throws errors compiling Qt core code (https://forum.qt.io/topic/140793/build-qt5-15-on-ubuntu-22-04-with-gcc-11; https://forum.qt.io/topic/139626/unable-to-build-static-version-of-qt-5-15-2/15). Options are presumably:

* force use of gcc-9
* upgrade Qt 5 beyond this incompatibility (e.g. from 5.12.4 to 5.15.3?)

Not sure which is best here.

RudolfCardinal:
Very happy for you to merge! Tests look good.

PR #326, opened 2023-12-22
Remove invalid assertion following SQLAlchemy 1.4 upgrade

martinburchell:
I missed this when upgrading to SQL Alchemy 1.4

Since SQLAlchemy 1.4, the return value from Session.execute() is now a CursorResult object and the type of
CursorResult.inserted_primary_key is now a tuple, not a list.

@RudolfCardinal the assert was added in 04599ca928a5d66da732598c55503986abda18da. I don't suppose you remember why? I suppose we could keep the `len(inserted_pks) == 1` if we wanted.

Old docs:
https://docs.sqlalchemy.org/en/13/core/connections.html#sqlalchemy.engine.ResultProxy.inserted_primary_key
New docs:
https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.CursorResult.inserted_primary_key

### DIFF
--- a/server/camcops_server/cc_modules/client_api.py
+++ b/server/camcops_server/cc_modules/client_api.py
@@ -1687,8 +1687,25 @@ def insert_record(
     rp = req.dbsession.execute(
         table.insert().values(valuedict)
     )  # type: CursorResult
-    inserted_pks = rp.inserted_primary_key
-    return inserted_pks[0]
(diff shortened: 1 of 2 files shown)


RudolfCardinal:
Have commented -- yes, likely a duff previous assert! Thank you!

RudolfCardinal (approved):

PR #344, opened 2024-04-04
Enable network log on failure / Update Danish / Tool to check missing client translations 

### LINKED ISSUE #336: Make clearer how to obtain verbose output upon network error
RNC:

- User received error message "There was a problem sending your completed tasks to the server"; this is from CamcopsApp::uploadFailed. No other info given ...

MB:

- It is possible to turn on verbose logging in single user mode. You have to jump through a number of settings screens to get there (Settings and Advanced Settings if I remember correctly).

Code relates to ``m_silent`` in ``NetworkManager``, ultimately related to ``CamcopsApp::isSingleUserMode()``.

Possibilities to make this more obvious to users (or could do several):

- Add this to docs.
- Put instructions within the short-form dialogue on how to switch to verbose mode.
- [Probably not: operate a "shadow" detailed logbox and offer to show it upon failure.]

martinburchell:
Fixes #336 . In single user mode, if there is a network failure there is now an option to retry with the error log displayed.

I've also updated the Danish. mostly by comparison with existing translations and with Google Translate. Please check!

In addition there's a new option to the `build_client_translations.py` script to report any missing translations and accompanying GitHub nag.


martinburchell:
@RudolfCardinal wrote this on the original ticket:

> Looks great. Two trivial things:

> * ``bool try_again_with_log`` could be ``const bool ...``
> * Task titles in translations use for format "Original [translation]" for the avoidance of doubt and the preservation of acronyms; therefore ``Compulsive Exercise Test`` should translate to ``Compulsive Exercise Test [kompulsiv træningstest]``

> and you've fixed some of my Danish too; thank you! I think that's it; otherwise good to go.

PR #351, opened 2024-04-30
Fix undefined behaviour when a task in a chain is prohibited and a subwindow is closed

### LINKED ISSUE #350: Incorrect questionnaires are being displayed during task chains
From CPFT AEDS:

> We've noticed that the incorrect questionnaires are being displayed sometimes during task chains (e.g. if you look at the most recent two patients on the server, the EDE-Q popped up in the middle of the COIIN task chain just now, and earlier today when ---- was getting a patient to do the clinical task chain they were somehow presented with the CHI-T and SUPPS-P).

Some thoughts:

* These task chains are set up with creation method "on demand or abort" so if the task chain was aborted, all of the remaining tasks would have been created. 
* What happens if one task chain is aborted and a different task chain is started?
* ISAAQ-10 prohibits commercial use. What happens if the user hasn't filled in their IP settings?

martinburchell:
Fix uninitialised flag in `TaskChain` class

Fixes #350 (or part of it). If a task chain had been aborted because of a missing IP setting, pressing the back button would cause the tasks up to the prohibited one to be displayed.

This probably would also have failed in the same way if a patient hadn't been selected and the task chain required it.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -3900,3 +3900,7 @@ Current C++/SQLite client, Python/SQLAlchemy server
   has forgotten their password. Fix a bug where if the initial password dialog was
   aborted, the next attempt to set up a password would fail.
   https://github.com/ucam-department-of-psychiatry/camcops/issues/346
+
+- Fix undefined behaviour if a task in a taskchain was aborted due to e.g. a missing
+  IP setting. Sometimes the tasks would be displayed if the back button was pressed.
+  https://github.com/ucam-department-of-psychiatry/camcops/issues/350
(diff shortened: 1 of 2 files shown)


RudolfCardinal:
! Yes, definitely agree with initializing that! Oops. Multiple chains start to exist when the menu is created, so this is an important signal hook (perhaps not the best design). Change looks good. For additional safety, do you think it's worth moving ``m_proceed_when_app_has_closed_last_task = false;`` to the start of ``TaskChain::startNextTask()``? That means resetting that flag also in case something like widget creation fails later in that function.

martinburchell:
> For additional safety, do you think it's worth moving `m_proceed_when_app_has_closed_last_task = false;` to the start of `TaskChain::startNextTask()`? That means resetting that flag also in case something like widget creation fails later in that function.

Yes good idea. Done.

PR #358, opened 2024-06-08
Use our own custom Qt build for running Qt tests

martinburchell:
Previously we were using a third party distribution of Qt for running tests. This is fine for some tests but it is useful to have the other third party dependencies for e.g. SQL Cipher database tests. It's also more likely to flag problems with our Qt etc builds. While GitHub supports uploading and downloading of artefacts that can be shared between jobs they expire after 90 days. So instead I've (ab)used the release mechanism so that Qt is built as an artefact of a draft CamCOPS release (along with the other third party components). We'll need to clean up old builds manually.


martinburchell:
This needs a bit more work so that the release gets uniquely tagged in some way.

RudolfCardinal (approved):

PR #362, opened 2024-06-25
Reformat all of the C++ with clang-format

martinburchell:
I've reformatted all of the C++ source with the `clang_format_camcops.py` script. I've also added a GitHub workflow to check. What do you think of the changes? 

I've changed the script to ensure version 14 of clang-format is being used but I don't know if running with a later or earlier version makes any difference.


martinburchell:
> The pygments C++ lexer is having problems with the changed `widgets/heightforlistwidget.cpp` in `HeightForWidthListWidget::event()` because of the split `case QEvent::Resize:` and `case QEvent::LayoutRequest:`. I'm inclined to agree that it looks a bit odd!

Hopefully fixed by moving inline comments to the next line.

RudolfCardinal:
Looks like a nice style. I presume this is a change that's tricky to apply in one go when other branches are active, because of the widespread effects? Small queries:

* #ifdef indentation seems inconsistent, e.g. it adds (nice) indents to ``gui_defines.h`` but removes them from ``platform.cpp`` -- ah, no, perhaps it's not inconsistent, and it will nest preprocessor definitions but not allow them to affect C++ code; that's fine. I see we've commented on this before in ``clang_format_camcops.yaml``.
* It splits single-line if statements without adding braces, e.g. ``Phq9::severity()``. I am a fan of "always braces" for clarity. I think this is a new option in clang-format-15 (``InsertBraces``, per https://stackoverflow.com/questions/26111162/ and https://clang.llvm.org/docs/ClangFormatStyleOptions.html); can we push to v15 and use this? I've attempted that.
* There are some imperfections, e.g. around ``const QString DEMO_SOUND_URL_1``, ``DEMO_SOUND_URL_2``, ``SOUND_COUNTDOWN_FINISHED`` in ``uiconst.cpp``.
* It also has a tendency to wrap definitions because of trailing comments (rather than move the comments). I tried to address that, per https://stackoverflow.com/questions/44731978/, with ``ReflowComments: true`` and adjusting some of the penalties. That makes it do some silly narrow-wrapped comments down the RHS, but that's easier to spot/fix as a human and disrupts the code less. However, it does break a number of code examples in longer comments.
* Comments are our job to sort anyway, I guess, so if they're making the wrap look silly, we can sort.
* Happy to proceed if you are. But pick the non-C++ changes and apply them after merging in other current work? Not sure how this is best done.

Files touched (ignoring the changes to C++ code):
* ``clang_format_camcops.py``
* ``clang_format_camcops.yaml``

martinburchell:
I'll leave this until after the next release in case we break something inadvertently.

martinburchell:
@RudolfCardinal I've rerun the formatting script having merged the comments changes. Unfortunately any inline comments that had been changed from:
```
    int foo = 0; // very long inline comment going over the maximum width of 80 characters
````

to:

```
    int foo = 0; 
        // very long inline comment going over the maximum width of 80 characters
```

are now:

```
    int foo = 0; 
    // very long inline comment going over the maximum width of 80 characters
```

So the association with the previous line is lost. 

Some possible options:
* Tell clang-format not to touch comments at all. This might mean we'd need to manually reformat quite a bit in the future.
* Move the comments above and aligned with the start of the line to which they were previously inline.
* Put some kind of continuation marker e.g. `// ... very long inline comment...`

What do you think?

RudolfCardinal:
Hmm. I thought this would be part of ``ReflowComments`` but that seems to be off already. I am generally keen on "code -> explanatory comment" for these sorts of short snippets, rather than having to rejig to have the explanation come first. The additional indent perhaps makes the code flow easier to see -- though these days we also have colour for that in many contexts. Part of the attraction of the automatic style is not having to think too much. And I think quite a lot of these comments have the leading ellipsis anyway, and that does make it visually to see the connection. So while I don't find this an easy decision I lean towards the leading ``...`` and letting it do its thing. What about you?

martinburchell:
> So while I don't find this an easy decision I lean towards the leading `...` and letting it do its thing. What about you?

Yes that was what I was thinking too. It seems that all of the clang-format options relating to comments control what is in the comments and not the indentation.

martinburchell:
@RudolfCardinal I've spot-checked a few of the files and as you say, most of the indented comments have the leading ellipsis already. I propose I don't go through them all again for this PR and we amend them later if we feel they need changing.

RudolfCardinal:
Great; I agree.

PR #379, opened 2024-10-15
Refactor python tests to use test factories

martinburchell:
Replaces helper methods in our `unittest.TestCase` subclasses with Factory Boy test factories. This should reduce the amount of model setup required for each test. Now only the fields being tested on a particular model need to be set as the factory should populate each model with sensible defaults.

The only real non-test code change here is the `UserGroupMembership.__init__()`, which seemed not to be necessary and was preventing the corresponding `UserGroupMembershipFactory` from being passed arbitrary arguments


RudolfCardinal (approved):
Re `UserGroupMembership.__init__()`: Ah, yes, I see the default SQLAlchemy constructor does sensible things with named arguments anyway (https://docs.sqlalchemy.org/en/20/orm/mapping_styles.html#default-constructor). Thank you!

PR #385, opened 2025-01-31
Update for SQL Alchemy 2

martinburchell:


martinburchell:
Back into draft while I address a problem with` joinedload()`

RudolfCardinal:
Thank you very much! Am working on this. Minor thing so far: e.g. in hamd.py, we have

  col = getattr(self.__class__, q)  # type: MappedColumn

and then use ``col.info``; this isn't defined in ``MappedColumn`` as far as PyCharm can see, so PyCharm flags it as a warning. I think PyCharm is wrong here but just wanted to check.

martinburchell:
> Thank you very much! Am working on this. Minor thing so far: e.g. in hamd.py, we have
> 
> col = getattr(self.**class**, q) # type: MappedColumn
> 
> and then use `col.info`; this isn't defined in `MappedColumn` as far as PyCharm can see, so PyCharm flags it as a warning. I think PyCharm is wrong here but just wanted to check.

PyCharm is right here. 

* These columns were added in `add_multiple_columns()`, which creates `Column` objects, either directly or via the `camcops_column()` function.  
* The `camcops_column()` function replaces the `CamcopsColumn` class, storing the extra information in the `info` dictionary.
* We can't use `mapped_column()` here because we can't put the Python type information in the column declaration. So the type should be `Column`, not `MappedColumn`.

RudolfCardinal:
OK - thank you! Will update.

martinburchell:
@RudolfCardinal I've reverted 5879a1b9a2f1f9061bd1f510d24575dbc13e4db5

which changed the type of the `horizontal_lines` argument to `TrackerInfo()` from:

(a list of floats) or (a list of ints) 

to:  

a list of (floats or ints)

MyPy and PyCharm may be in conflict with each other here. MyPy doesn't like something typed as a list of ints being passed in here. I think if we want horizontal lines to accept a mixture of floats and ints we'd need to change the signature to be `Optional[Union[list[float], list[int], list[Union[int, float]]]`. I have not tested this.

RudolfCardinal:
Shall we try that? It's possible that e.g. [5, 7.5, 10] is used somewhere.

Issue #366, opened 2024-06-28
It is possible to enter zero height in the BMI task


martinburchell:
I've not investigated this fully but it looks like someone has managed to enter zero height in the BMI task and this has caused the server to throw an exception when dividing by zero. It looks like the client might crash as well.

martinburchell:
The Linux client displays NaN for zero height. Might be the same on other platforms but we could guard against any possible crash by setting an unrealistic low minimum value.

RudolfCardinal:
Agree entirely with modification to server. ``bmi.py``.

In the client calculations, I think it's not guaranteed that the client couldn't crash, but almost guaranteed (perhaps compiler convention rather than absolute standard? e.g. https://stackoverflow.com/questions/4745311/c-division-by-0). So perhaps sensible to have a check in Bmi::bmiVariant() that returns QVariant() if height_m <= 0.0. If it's displaying NaN rather than "?", then I presume it's explicitly representing a NaN rather than a null QVariant(), so some of the other code might misbehave. But this code should cope with zero.

Then to avoid zeros getting in also -- yes, maybe; presumably would need a "minimum" parameter for QuHeight? For the Bmi task (and any others using this in practice) I imagine 0.1 m might be a sensible minimum.

I wonder why this happened, though; perhaps someone was insistent on using the BMI task to measure weight and didn't know height. In these circumstances, it might be more desirable to allow 0, producing a null BMI that is appropriately handled, than to enforce something duff (producing a numerically valid but nonsensical BMI). I suspect this is a realistic clinical scenario so perhaps we should roll with it. What do you think?

martinburchell:
As per offline discussion, we'll still allow zero height but guard against it and look towards allowing patients to opt-out of tasks they do not wish to complete.

PR #387, opened 2025-02-25
Replace use of deprecated Column.copy() prior to SQLAlchemy 2 upgrade

martinburchell:
`Column.copy()` and `ForeignKey.copy()` are removed in SQLAlchemy 2.0, well strictly speaking, renamed to `xxx._copy()` to indicate they aren't part of the API.

The recommended replacement is` Table.to_metadata()`, which sort of does what we want except that we can't exclude columns and we don't want the `ForeignKey` constraints. See discussion at https://github.com/sqlalchemy/sqlalchemy/discussions/10618 where excluding columns was proposed as a feature but rejected.

So we use `Table.to_metadata()` to do most of the work and then remove any `ForeignKey` definitions by altering the columns. If we are excluding a column within an included table this just gets changed to be an Integer and the columns 
are ignored when copying the rows.

One side effect of this is that the index property for foreign keys is no longer retained, meaning that generation of  the CRATE data dictionary in `cc_anon.py` no longer sets the `index` column for these. I'm not sure how useful this is anyway give that CRATE can generate its own data dictionary.

I've created an additional SQLite in-memory database for testing, enabled with the decorator `@pytest.mark.usefixtures("setup_dest_session")` and then available as `dest_session` and `dest_engine` on the test class. See `cc_dump_tests.py` for an example. Unlike the main test database, this does not create tables at the start of the test run, assuming the tests will do this themselves. Any tables are dropped at the end of each test so it could potentially be quite slow.

In testing SQLite dump from the web interface, I've fixed a couple of bugs, namely:

* More columns from INFORMATION_SCHEMA need to be overridden (possibly broken since MySQL 8.0?)
* A fix for invalid BMI when exporting SNOMED codes (following on from #366).


RudolfCardinal:
Looks good. But the actual dump to XLSX or ODS was failing - one re timezone-aware datetime values (causing an error in openpyxl), and the other about a string subclass (failing to get through a type-based mapping in pyexcel-ods3). Fixed in cardinal_pythonlib, https://github.com/RudolfCardinal/pythonlib/pull/30. Version string for that library may need resolution.

PR #388, opened 2025-03-11
CIS-R phrasing

RudolfCardinal:
Ensure caveats about CIS-R use are prominent. Also fix a trivial display bug for the GP_YEAR question (valid database value of 0 was shown as "?" but should have been "None", because for most other questions, 0 is a non-response code).


RudolfCardinal:
Hmm. Unusual build error; hash mismatch, for pip itself. Try again later?

martinburchell:
> Hmm. Unusual build error; hash mismatch, for pip itself. Try again later?

Yes given the other jobs succeeded sounds like a one-off. You should be able to re-run a job manually from the failure page. I know that Ubuntu 20.04 will no longer be supported for workflows soon but this doesn't look like that is the problem.

RudolfCardinal:
Thanks - yes, succeeded on retry.

tablet_qt/tasks/cisr.cpp
```diff
@@ -3625,25 +3628,29 @@ QVector<QString> Cisr::panicSymptomFieldnames() const
     return fieldnames;
 }
```
martinburchell, line 3630: I'd delete the commented-out code or disable it with a preprocessor directive if it's likely to be needed in the future.

tablet_qt/tasks/cisr.h
```diff
@@ -430,8 +431,8 @@ class Cisr : public Task
     QVector<QString> panicSymptomFieldnames() const;
     CisrQuestion nextQ(CisrQuestion q, CisrResult& getResult) const;
     CisrResult getResult() const;
-    QString diagnosisNameLong(int diagnosis_code) const;
-    QString diagnosisReason(int diagnosis_code) const;
+    // QString diagnosisNameLong(int diagnosis_code) const;
```
martinburchell, line 434: As above

server/camcops_server/translations/camcops_translations.pot
```diff
@@ -1,14 +1,14 @@
 # Translations template for CamCOPS.
-# Copyright (C) 2024 Rudolf Cardinal
+# Copyright (C) 2025 Rudolf Cardinal
```
martinburchell, line 2: Should it be (C) University of Cambridge now as for the code?

martinburchell (changes requested):
See comments. Otherwise looks good!

Unrelated to this change I notice that the missing migrations workflow is silently failing following the changes to how the alembic config picks up the Database URL. I'm happy to fix that on my own branch but any pointers would be helpful!

RudolfCardinal:
Alembic problem fixed, I think, in https://github.com/RudolfCardinal/pythonlib/pull/32.

RudolfCardinal:
Will likely fail until https://github.com/RudolfCardinal/pythonlib/pull/32 merged and pushed to PyPI, but I think not for other reasons.

PR #390, opened 2025-04-11
Warn the user if multiple server-created patients exist with the same ID, group etc

### LINKED ISSUE #375: It should not be possible to create the same patient twice in the same group on the server
We should enforce this either at the user interface level or the database level, or both.

martinburchell:
Closes #375

Enforcing unique server-created patients at the database level or indeed at the form level is hard so instead I've opted to continue to allow this (CamCOPS *should* cope OK) but there are now big red/pink warnings if two server-created patients exist with the same identifier in the same group.

I've also tweaked the big red/pink warnings to be a lighter background colour so that they now pass the WCAG AA contrast test.


RudolfCardinal (approved):
Looks good. One efficiency point:

    duplicates = False
    for patient in page:
        if any([idobj.duplicates for idobj in patient.idnums]):
           duplicates = True

perhaps could be

    duplicates = any(idobj.duplicates for patient in page for idobj in patient.idnums)

... removing (a) the loop through patients beyond the first hit, and (b) I'm not sure whether the conversion from generator to list is required (but I might be missing something there).

PR #395, opened 2025-05-23
Upgrade Qt to 6.5.5

### LINKED ISSUE #308: The About Qt dialog is blank on iOS / Qt6.5

martinburchell:
* Update Qt to the latest 6.5 LTS LGPL version (6.5.5). The git tag for this is `6.5.5-lts-lgpl`, which makes the version file parsing/comparison code a little trickier. `qt_version.txt`now holds the git tag.
* This fixes #308. We also have to make sure the dialog doesn't inherit the style sheet from the parent, otherwise the Dialog is displayed as non-native and again appears incorrectly on iOS. 
* The [Android camera crash bug](https://bugreports.qt.io/browse/QTBUG-116526) is fixed so we can remove our patch for that.
* There is [a new bug where QLineEdits in Dialogs do not have the focus](https://bugreports.qt.io/browse/QTBUG-125337) so there's a new workaround for that.
* The [static FFmpeg linking bug](https://bugreports.qt.io/browse/QTBUG-115052) is still not fixed so there's an updated patch for that.
* Use our own build of Qt for the missing client translations workflow. aqtinstall does not have a 6.5.5 build and this makes things consistent with the C++ tests workflow.
* Type hints for `build_qt.py` cherry-picked from the `sqlalchemy2` branch
* Remove/update Ubuntu 20.04 workflows
* Security updates: asteval, tornado and safety dependencies.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -3953,5 +3953,10 @@ Current C++/SQLite client, Python/SQLAlchemy server
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Update the Docker image to use Debian 11. Debian 10 has now reached end-of-life.
+
 - **Minimum Python version now Python 3.9.** Python 3.11 and 3.12 supported.
+
+- Fix cursor placement when entering the access key on iOS. The workaround for
+  https://bugreports.qt.io/browse/QTBUG-115756 is now only applied for Android.
+
 - Qt version now 6.5.5.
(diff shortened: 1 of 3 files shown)


martinburchell:
@RudolfCardinal Sorry I merged this accidentally into master instead of the other way round! Would you mind checking the changes and putting any comments here? Thanks!

martinburchell:
> Sorry, I've lost the sequence - I think this is done - OK to delete the branch?

Yes if you're happy with the changes

PR #401, opened 2025-07-14
Provide suggestions for debugging '403 Forbidden' messages

RudolfCardinal:
Add to docs re "403 Forbidden" errors and tracking them down.

### DIFF
--- a/docs/source/administrator/server_troubleshooting.rst
+++ b/docs/source/administrator/server_troubleshooting.rst
@@ -355,8 +355,78 @@ Fix for Azure Application Gateway:
 - https://docs.microsoft.com/en-us/azure/application-gateway/rewrite-http-headers
 
 
-Tablet upload errors
---------------------
+App (phone/tablet device) errors during registration or upload
+--------------------------------------------------------------
+
+"403 Forbidden"
+~~~~~~~~~~~~~~~
+
+**The CamCOPS app (on a phone/tablet device) fails with an error including
+“403 Forbidden”, or “server replied: Forbidden”.**
+
+This error suggests that something is preventing the mobile device from talking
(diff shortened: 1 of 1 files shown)


docs/source/administrator/server_troubleshooting.rst
```diff
@@ -355,8 +355,78 @@ Fix for Azure Application Gateway:
...
+
+- It's possible to get combination problems, e.g. a firewall not allowing POST
+  requests from a specific address range.
+
+You may get a better view of the problem by running this command from a Linux
+machine:
+
+.. code-block:: bash
+
+    curl -i -X POST -H "some_header" -d "some_body" https://<CAMCOPS_SERVER>:443/database
```
martinburchell, line 401:
```suggestion
    curl -i -X POST -H "some_header" -d "some_body" https://<CAMCOPS_SERVER>:443/api
```

I think we decided to use ` /api` in favour of `/database` but with an alias for the latter for backwards compatibility.

If the server and firewall are working properly and someone literally enters `-H "some_header" -d "some_body"`, they will get a 500 error. [There is an issue for this already](https://github.com/ucam-department-of-psychiatry/camcops/issues/323). I expect most administrators will struggle to know what a valid request looks like and maybe a future development would be a simple ping operation? As it stands, it is useful to have this documentation for ourselves if nobody else. Getting the 500 instead of the 403 indicates that the firewall has allowed the request through.

martinburchell (changes requested):

RudolfCardinal:
Thanks. Good point. I'll try to fix https://github.com/ucam-department-of-psychiatry/camcops/issues/323 too.

martinburchell:
Thanks! I added an automated test. Looks good to me.

PR #404, opened 2025-07-21
Don't treat warnings from Eigen as errors and client build workflow

martinburchell:
See https://gitlab.com/libeigen/eigen/-/issues/2787 

Because we are treating warnings as errors the client build fails. 

Also a workflow to build the client on Ubuntu 22.04 using pre-built Qt release.

### DIFF
--- a/.github/scripts/build_client.sh
+++ b/.github/scripts/build_client.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
+echo "Building client..."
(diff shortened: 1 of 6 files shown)


RudolfCardinal (approved):
Looks good (though I've not tested; I think that via the workflow). Are the Boost/mlpack/armadillo bits no longer necessary? (If so: good.) And ?over-indentation at camcops.pro:257 (cosmetic).

Issue #410, opened 2025-08-07
Validation of server URL on Samsung Galaxy Tab A7 is inconsistent


martinburchell:
In the process of fixing #407 I have noticed that the validation of the server URL on the Samsung Galaxy Tab A7 waits for a whole "word" to be entered before running validation, even with predictive text disabled.

So if I am entering:
`https://camcops.example.com/api`

I don't see any validation until I enter the colon after the https. Then it says **Invalid** when entering `://camcops` until the dot, when it says **Valid**. Then it continues to say **Invalid** until I enter the next dot. Finally it will only say **Valid** after entering the forward slash at the end. 

Next if you start deleting characters, it says **Valid** until you have deleted everything after the `https://`, after which it says **Invalid** again.

This is a known problem documented at https://bugreports.qt.io/browse/QTBUG-37533, which was closed because the problem is with Samsung and there is a workaround, to add this line before the on-screen keyboard is displayed:

`qputenv("QT_ANDROID_ENABLE_WORKAROUND_TO_DISABLE_PREDICTIVE_TEXT", "1");`

but we would need to run this only when the widget gets the focus and disable it when the widget loses the focus otherwise there are unintended side effects for other widgets. I tried just adding it before and after displaying the patient registration dialog but that then also ignored the flag to turn off auto upper case for the access key.

PR #412, opened 2025-08-08
Replace custom strict number validators with standard Qt Validators and user feedback

### LINKED ISSUE #407: Can't enter decimal point in demo questionnaire (double, range -0.05 to -0.09, 2dp)

martinburchell:
Fixes #407 and #410.

### Summary

I have modified `QuLineEdit` and derived classes to use `ValidatingLineEdit` (previously only used by `PatientRegistrationDialog`). This provides visual feedback for valid/invalid input: a red or green background accompanied by a **Valid** or **Invalid** label. The label appears to the right of the text box unless configured to appear below it (`vertical` flag). In this context **Invalid** is really `QValidator::State::Intemediate` i.e. something that isn't currently valid but might be if the user typed in a few more characters. Input that could never be valid (`QValidator::State::Invalid`) isn't allowed at all in the underlying `QLineEdit` and doesn't result in the `textChanged` signal that we are listening for being sent.

I've moved quite a bit of the `QuLineEdit` functionality into `ValidatingLineEdit`. The red/green colours are now set in CSS rather than using `QPalette` to be consistent with the other widget colours.

It seemed sensible to change  `ValidatingLineEdit` from being a layout to a widget. I've made access to the underlying `QLineEdit` protected so there are a group of methods on `ValidatingLineEdit` that are just wrappers to `QLineEdit` methods. I'm not sure if this was a good design decision or not! I've replaced `extraLineEditCreation()` on
`QuLineEdit` subclasses with specific methods `getValidator()` and `getInputMethodHints()`.

I've replaced the strict integer/double validators with the standard Qt ones and added validators for 64-bit signed and unsigned integers based on `QIntValidator`. The `allow_empty` check is now done in `ValidatingLineEdit` rather than the validators themselves.

I've moved trimming whitespace to the `ValidatingLineEdit` subclasses. I'm not sure where it might be desirable _not_ to trim whitespace but it didn't seem right to be doing this in `PatientRegistrationDialog` or in `ValidatingLineEdit` itself.

`QuLineEditDouble` now cares about the order of its min/max range arguments.

### Testing

A lot more automated tests. `QuLineEdit` changes tested manually because it is difficult to write automated tests for these (the `Questionnaire` class is currently dependent on `CamcopsApp` and that will pull in everything else).

All tasks/forms with `QuLineEdit` subclasses and valid ranges tested.

### Also in this PR

* Ignore build directories named `build-*` when generating code documentation
* Security fix for PyPDF
* Treat warnings as errors when compiling automated tests


RudolfCardinal (commented):
Looks excellent; thank you! The new behaviour similarly seizes on "happy" input, e.g. for range 13-19, if you type "14" then delete the "4", it reverts to 14 on focus exit, but that seems fine. It seems to work very nicely and be much less complicated (and therefore: more reliable) than before. I've not tested the proquint version but I see there are tests for that.

PR #424, opened 2026-01-22
Better handling of incomplete patient registration due to firewall blocking genuine requests

### LINKED ISSUE #367: 2 patients with clashing ID numbers in single user mode
User reports not being able to upload in single user mode. The error log (see screenshot) has the error:

`Failure: 2 patient(s) having clashing ID numbers`

Image

martinburchell:
Fixes #367 

When patient registration fails, delete any partially registered patients from the client database. Before this change a second patient record would be created the next time the patient registered.

The changes on the server are just for debugging. It may be useful to simulate a Forbidden response code if we encounter similar problems in the future.

Also includes another round of security fixes.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -4020,3 +4020,7 @@ Current C++/SQLite client, Python/SQLAlchemy server
   https://github.com/ucam-department-of-psychiatry/camcops/issues/415
 
 - **Minimum Python version now Python 3.10.**. Python 3.9 has reached end-of-life.
+
+- Fix bug where it was not possible for a single user to upload tasks because of
+  an incomplete patient registration.
+  https://github.com/ucam-department-of-psychiatry/camcops/issues/367
--- a/server/camcops_server/cc_modules/cc_client_api_core.py
+++ b/server/camcops_server/cc_modules/cc_client_api_core.py
@@ -197,6 +197,14 @@ class ServerErrorException(Exception):
     pass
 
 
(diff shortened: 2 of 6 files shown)


RudolfCardinal (commented):
Looks good. The only thing that popped out for me is the deletion of all records from the patient table. Are there any conditions in which this might be dangerous? e.g.
* If a user switched from clinician mode to single-user mode, are we sure there will be no other patients? (Set partly via ``setDefaultPatient()``?)
* Is it possible for there to be existing task records that would rely on the patient record at this point?

If no to both, all looks good to me. Probably ``registerPatientWithServer()`` does this cleanup first?

martinburchell:
> Looks good. The only thing that popped out for me is the deletion of all records from the patient table. Are there any conditions in which this might be dangerous? e.g.
> 
>     * If a user switched from clinician mode to single-user mode, are we sure there will be no other patients? (Set partly via `setDefaultPatient()`?)
>     * Is it possible for there to be existing task records that would rely on the patient record at this point?

Good points.

We clear pretty much everything on mode change including patients:

https://github.com/ucam-department-of-psychiatry/camcops/blob/baad31a120d052ef5c644c230f9f542cfa30f5f1/tablet_qt/core/camcopsapp.cpp#L303

For a brand new registration there would be no tasks. Any patient at this point would only be the partly registered one. Re-registering should delete the currently "selected" patient along with any tasks, but when it was failing part way through, the partly registered patient would not yet be selected and nothing would get deleted.

PR #435, opened 2026-07-28
Replace wkhtmltopdf and PDFkit with Weasyprint

### LINKED ISSUE #383: wkhtmltopdf going end-of-life
Will need some refactoring soon; see https://github.com/RudolfCardinal/pythonlib/issues/22

martinburchell:
Fixes #383

* Remove wkhtmltopdf from the Docker image
* Add page numbers for Weasyprint in a style sheet and pass that to Weasyprint
* Remove IMG fallbacks where SVG isn't supported. Weasyprint supports SVG now, as do all of the major browsers. It appears to be broken if the PDF engine is changed to xhtml2pdf and we can spend time fixing that if Weasyprint ends up not being an option. 
* Replace obsolete table heading width attributes with styled COL elements, which Weasyprint supports 

For now PDFKit will still work if the package is installed manually and enabled in `cc_constants.py`. We might want to consider removing it completely in the future.


RudolfCardinal:
Very nice - thank you! Excellent also that we can remove the PNG fallback for SVG.

RudolfCardinal (approved):

